PR #1, opened 2018-04-04
(feat): nnnnat reaction theme

nnnnat:
## Issue
Material UI doesn't use the Reaction styles

## Solution
Material UI has several options for theming the UI lib. This is just a first pass at adding the Reaction colors and type styles to the theme.

## Testing
1. Start up the app and see the font family is now Source Sans Pro and the Primary, Secondary and Error theme options now use colors from the RC [style guide](https://zpl.io/aBO3D8A).

## Notes
Once we get some larger UI built out @cassytaylor and @rymorgan should review and make suggestions on which theme variables we replace with RC styles.

### DIFF
--- a/components/Header.js
+++ b/components/Header.js
@@ -6,9 +6,9 @@ import Typography from "material-ui/Typography";
 class Header extends Component {
   render() {
     return (
-      <AppBar position="static" color="primary">
+      <AppBar position="static">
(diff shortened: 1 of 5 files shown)


willopez:
Updates pulled into master, thank Nat.

PR #2, opened 2018-04-04
(feat): storefront using reaction docker base

nnnnat:
## Issue
Needed the Storefront example application to run within the Reaction Commerce Docker setup. Also needed a way for the Storefront container to communicate with the GraphQl service container.

## Solution
Added all the necessary docker config files from the Reaction Base.
Created a network definition that both containers can communicate on.

## Testing
Follow the setup steps within the [README](https://github.com/reactioncommerce/reaction-next-starterkit/blob/7a1ea0fa0a15295d1f9ca817fa60bc6182e4c18b/README.md). 
The Storefront app should start and run at localhost:4000 with no errors in the terminal.

## Notes
You'll need to run the Storefront app with `docker-compose up -d && docker-compose logs -f web` to see any terminal logs.


nnnnat:
@aldeed  @ticean  we were able to get this docker setup working but one issue we're having is the `yarn install` doesn't seem to install anything when running `docker-compose up` or `docker-compose up --build`. If we run `yarn install` locally before starting the container everything works fine. Not sure if I'm setting the build env correctly.

src/lib/apollo/initApollo.js
```diff
@@ -14,7 +14,8 @@ const create = (initialState) =>
     connectToDevTools: process.browser,
     ssrMode: !process.browser,
     link: new HttpLink({
-      uri: "http://localhost:3030/graphql-alpha",
+      uri: `${process.env.PROTOCOL}://${process.env.HOST}:${process.env.PORT}/${process.env
+       .ENDPOINT}`,
```
aldeed: I would destructure these (including METEOR_TOKEN) from `process.env` toward the top of the file and check each one throwing a specific error if it isn't set. That way when someone inevitably starts it up with improper configuration they'll get a more helpful error message like "You must set PROTOCOL env variable".

I'm also not sure why it would be necessary to break this up rather than just having a single `GRAPHQL_URL` variable.

PR #31, opened 2018-04-06
feat: Add gql file loader

willopez:


willopez:
@mikemurray I have added the configuration necessary to load queries from .gql files.

PR #40, opened 2018-04-12
Tag data fetching for navigation

mikemurray:
Issue: #10
Impact: **minor**
Type: **feat**

## Issue

Create a way to fetch tags for the navigation bar.

## Solution

- Create a HOC to get the primary shop ID, and another HOC to get the tags based on that `shopId`. - - Recursively fetches more tags if more is available.
- Updated apollo config and initialization to support client-side data fetching

## Testing

**Pre-req**

You'll a `src/browser.config.js` file with the client URL to the GraphQL endpoint (most likely `http://localhost:3030/graphql-alpha` ) along with the meteor login token.  see `src/browser.config.example.js`.

You will still need the .env file for docker if you're using.

1. Start the `starterkit` and `reaction` docker containers (you may need a meteor login token from reaction for the starter kit to be able to connect to the GQL endpoint)
2. Once the reaction is started, navigation to `localhost:4000` to view the starter kit. See tags populated
3. With the dev-tools in Reaction, populate with the medium dataset
4. Refresh starter kit and you should see navigation items, with sub-items and sub-sub items

## Other considerations

The desktop navigation starts wrapping items when there are too many. We need to find a solution to scroll the navbar items when this happens. The [Material UI Tabs](https://material-ui-next.com/demos/tabs/) have this ability, so we could try with that component.


willopez:
@nnnnat it looks good on my machine, lets merge it

spencern:
@mikemurray did you create an issue for scrolling the navbar if it's overflowed?

Issue #43, opened 2018-04-16
Implement Token-based login for dev


aldeed:
## Task

**Note**: this is a token based login for use in development only.

To make testing auth a little nicer as an interim solution until we have a true login flow, do the following:

1. Add a "Login" button in the nav bar.
1. Clicking "Login" opens a modal. Modal contains a “Meteor token” text field and Save and Cancel buttons. Use Material Dialog, TextField, Button, etc. components.
1. When you click Save on the modal, save the entered token to mobx state and local storage.
1. Reload as that user. Easiest is probably to just hard refresh the page.

PassportJS could be a good solution here.

In server code, come up with a solution to get proper auth token from the `request` somehow. Might need to be from a cookie. Whatever solution this is needs to be in sync with the above, changed when the user saves a different token.

## Goal

- I can change who I'm logged in as (which token is used) while the app is running
- Server rendered page shows the correct content for me if I've previously "logged in" in the browser I'm using.

PR #45, opened 2018-04-17
feat: Token-based login for dev

mikemurray:
Ticket: #43 
Impact: **minor**
Type: **feat**

## Issue

Implement a token based authentication system

## Solution

- Added a login dropdown with an input field with a save button
- Used a Popover and omitted "Cancel" button as you cancel by clicking outside the popover.
- Added a mobx authStore for saving the token in state as well has saving / getting token from cookies
- Saved the token as a cookie and providing it to the client that way as well as storing it in mobx
- Didn't use Localstorage as it's redundant to the cookies
- Provide the token to apollo on both client and server
- Removed login token from env vars and various places where it was used from those vars

## Breaking

Login token must be supplied using the login dropdown.

## Testing

1. Start app
1. Use the Account icon in the navigation 
1. Enter a valid meteor login token and hit save
1. App should load with data


mikemurray:
Adding tests now, should be ready soon.

mikemurray:
Ready for review

Issue #22, opened 2018-04-05
Product Detail Basic Information


nnnnat:
@spencern commented on [Tue Apr 03 2018](https://github.com/reactioncommerce/reaction/issues/4130)

Add the basic information section to the Product Detail layout as defined in #4126 

![image](https://user-images.githubusercontent.com/1203639/38232925-1ef50dd8-36d6-11e8-8caa-b724f51a7527.png)

The product detail layout contains the product `price`, `vendor`, and `description`

PR #53, opened 2018-04-20
feat: PDP basic info

mikemurray:
Issue: #22 ( depends on #21 )
Impact: **minor**
Type: **feat**

## Issue

Create basic info on the PDP page with price, vendor, and description fields.

## Solution 

Create a component that can

- Display price, vendor, description if the data is passed to the component.

## Test

- Start the app and go to `http://localhost:4000/product/1234`
- See the following

reaction_shop


src/components/ProductDetailInfo/ProductDetailInfo.js
```diff
@@ -0,0 +1,43 @@
...
+import Grid from "material-ui/Grid";
+
+/**
+ * Product detail basic info fields
+ * @class ProductDetailTitle
+ */
+class ProductDetailInfo extends Component {
+  static propTypes = {
+    /**
+     * Prodcut description
```
willopez: Typo here, should be Product

src/components/ProductDetailInfo/ProductDetailInfo.js
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Product vendor
+     */
+    vendor: PropTypes.string
+  }
+  render() {
+    const { priceRange, vendor, description } = this.props;
+
+    // Render all props are undefined then skip rendering component
```
willopez: I believe this is a typo, it should be something like: "If all props are undefined ..."

src/pages/product.js
```diff
@@ -7,6 +7,9 @@ import withShop from "containers/shop/withShop";
 import Layout from "components/Layout";
 import ProductDetail from "components/ProductDetail";
 
+// TODO: Data will eventually come from GraphQL
+import sampleData from "components/ProductDetail/sampleData";
```
willopez: What are your thoughts on using a jest convention for naming mocked data. I like using a folder named `__mocks` that would contained a file name `product.js` with the product object? and we can use those mocks in jest tests.

mikemurray:
@will made those changes, also moved the sample data to `__mocks__` since it'll only be needed for tests once we hook the PDP page to real data.

spencern:
@mikemurray looks like you tagged the wrong @willopez in this issue :)

mikemurray:
Yup seems that way

PR #61, opened 2018-04-25
Add variant options list

willopez:
This PR will add components to display a variant's options list.

### Testing
1. Verify mocked variants render correctly on desktop/mobile
2. Click on an option and verify it's selected.


willopez:
@mikemurray comments addressed.

PR #75, opened 2018-05-01
fetch data for PDP

jshimko:
This PR relies on https://github.com/reactioncommerce/reaction/pull/4218 to work (which should hopefully be merged soon)

Updates:
- fetch product item data using URL slug or ID
- update any PDP components with mapped data
- update mock data to match GraphQL data

### DIFF
--- a/src/lib/apollo/initApolloServer.js
+++ b/src/lib/apollo/initApolloServer.js
@@ -17,13 +17,13 @@ const errorLink = onError(({ graphQLErrors, networkError }) => {
   if (graphQLErrors) {
     graphQLErrors.forEach(({ message, locations, path }) => {
       // eslint-disable-next-line no-console
-      console.log(`[GraphQL error]: Message: ${message}, Location: ${JSON.stringify(locations)}, Path: ${JSON.stringify(path)}`);
+      console.error(`[GraphQL error]: Message: ${message}, Location: ${JSON.stringify(locations)}, Path: ${JSON.stringify(path)}`);
     });
   }
 
   if (networkError) {
     // eslint-disable-next-line no-console
-    console.log(`[Network error]: ${networkError}`);
+    console.error(`[Network error]: ${networkError}`);
   }
 });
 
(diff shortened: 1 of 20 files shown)


spencern:
@jshimko How close are we to having a PR with this?

jshimko:
This is still depending on the catalog schema PR in the main Reaction repo (https://github.com/reactioncommerce/reaction/pull/4218), but it's ready to be tested/reviewed.  

```
reaction init -b refactor-4196-aldeed-catalog-schema-changes catalog-schema
cd catalog-schema
docker-compose up -d --build
```

Then start up this PR branch and everything should work.  You should see the PDP is now completely populated by the GraphQL API data.

src/lib/apollo/initApolloServer.js
```diff
@@ -11,17 +12,34 @@ if (!process.browser) {
   global.fetch = fetch;
 }
 
+// error logging for Apollo Link
+const errorLink = onError(({ graphQLErrors, networkError }) => {
+  if (graphQLErrors) {
+    graphQLErrors.forEach(({ message, locations, path }) => {
+      // eslint-disable-next-line no-console
+      console.log(`[GraphQL error]: Message: ${message}, Location: ${JSON.stringify(locations)}, Path: ${JSON.stringify(path)}`);
```
nnnnat: could we use the logger util for something like this?
https://github.com/reactioncommerce/reaction-next-starterkit/blob/master/src/lib/logger/index.js

jshimko: Well, maybe... but that logger is using `chalk` (for terminal colors) and these particular errors can happen on the server _and_ the client.  I think we probably need a more isomorphic solution for this use case.  We could certainly use [Bunyan](https://github.com/trentm/node-bunyan) for that.  Then we'd have all of the same API's we're used to in Reaction (for piping logs to other places, etc).

jshimko:
@nnnnat FYI, it's hidden in the Github UI now because I pushed changes, but I did respond to the logging questions.

src/lib/apollo/initApolloServer.js
nnnnat: yeah, I wasn't sure if our `lib/logger` would work in these cases. I'm good with just doing the `console.error` here for now and maybe we revisit our logger soon in another PR

nnnnat:
@spencern  @aldeed  I think this is good 2 go, but should we wait on [PR 4218](https://github.com/reactioncommerce/reaction/pull/4218) before we merge this one?

PR #77, opened 2018-05-02
Add inventory status badges to variants and options

willopez:
Add inventory status badges to variants/options lists. 

**NOTE**: The badge style will be updated when the designers finish the design of the PDP page.

### Testing
1. Add a new product with a variant and an option and publish the product.
2. By default the variant will have a backorder badge.
3. Verify a backorder badge is visible on the new product variant in the  grid and PDP.
4. Toggle the backorder switch to turn off backordering
5. Grid and PDP should display a "Sold Out" badge
6. Repeat steps 3-5 for the variant option.

### DIFF
--- a/src/components/ProductItem/ProductItem.js
+++ b/src/components/ProductItem/ProductItem.js
@@ -7,8 +7,8 @@ import Fade from "material-ui/transitions/Fade";
 import Hidden from "material-ui/Hidden";
 import Typography from "material-ui/Typography";
 import LoadingIcon from "mdi-material-ui/Loading";
-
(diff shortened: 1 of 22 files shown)


src/components/ProductItem/ProductItem.js
```diff
@@ -38,12 +38,12 @@ class ProductItem extends Component {
   }
 
   get productStatus() {
```
spencern: Let's work with @nnnnat and @aldeed to see if we can move all of this logic out of the client and publish an inventory status to the catalog instead.

Issue #65, opened 2018-04-26
Product grid image size issue


mikemurray:
## Issue

Product grid items can have different sizes. The height of the medium and large grid items seems to be affected by the image. The height of images should all remain the same, but their widths should be different based on their weights.

### The highlighted image below shoun't be tall...

I should be the same height as the item next to it, but just as wide as it is now.

![reaction_shop](https://user-images.githubusercontent.com/42217/39318356-c586d528-4932-11e8-9584-bd86d5b8c864.png)

## Possible Solution

Images need to `object-fit` cover.

kieckhafer:
@mikemurray what are your thoughts on the best way to go about this? Traditionally, I'd say set a fixed height (i.e. 325px), an then adjust that with media queries based on the width. Since we no longer have any stylesheets though, what's a good way to do this?

Issue #71, opened 2018-04-27
Product grid has no horizontal boundaries


spencern:
When viewing the product grid on a widescreen monitor, the product grid scales to the full width.

@rymorgan @cassytaylor what should the max width of the grid be? 

![image](https://user-images.githubusercontent.com/1203639/39371451-fa3b4840-49fe-11e8-96d9-98792d46998d.png)

rymorgan:
Researching trends for an answer to this and how it should work responsively.

rymorgan:
The short answer, for now, we should lock the grid at 1440 with padding on both sides.

**Further findings -** 

 -  The height of the images should be locked Almost no brands use   
   multiple width images on their standard grid only when they have special merchandised category or collections. 
 - The product image should be square it's up the brand to upload the right image only
   specialed brands use portrait or landscape
 - Images should fill grid w/ proper padding between them for the following scenarios - Desktop (4 wide), Tablet (3), Phone (1)
 - We can allow images to flex when the window shrinks before a breakpoints
 - In the future, we'd move filtering to a dropdown above the grid at vertical tablet size

rymorgan:
Anyone interested in the competitive research around product grids here are my findings. (click on Product grid - size and flex tab.) https://docs.google.com/spreadsheets/d/1_whIDhSjW8RZ0N5VqpKwTMMuuR-KaT_FUSN6yIx5zTY/edit#gid=63930977

nnnnat:
Resolved with PR #84

PR #84, opened 2018-05-03
fix: image grid sizing issue

kieckhafer:
Fixes Issues #65 & #71 

Product grid items have three different sizes. The heights of medium and large items were affecting the grid heights, as there were not stable. In addition, there was no container, so the grid was always full-width

This PR adds a fixed height to product grid images, to keep the grid... a grid. It also adds a 1440px container, as discussed in ticket #71, to contain the grid.

It uses `material-ui-next`s built in CSS-in-JS `sm`, `md`, and `lg` breakpoints.

The PR also adds some docs (a link to the actual Material UI docs).

This PR also removes leftover test code from #59

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,3 +61,7 @@ _**NOTE:** The above command is assuming ether the `devserver` or `reaction` is 
  - [MobX](https://mobx.js.org/getting-started.html)
  - [nextjs](https://github.com/zeit/next.js/)
  - [Material UI](https://material-ui-next.com/)
+
+ ## Reference links for development
+ ### CSS in JS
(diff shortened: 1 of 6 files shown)


kieckhafer:
@rymorgan just want to confirm that we aren't currently doing the three sizes on the grid, and we might save that for future development?

kieckhafer:
@nnnnat if we aren't using the three different sizes anymore, most of this PR is probably not useful.

We should probably keep the container style, but I'm thinking i'll probably get rid of all the style concerning the images in the grid.

Does that sound right?

nnnnat:
@kieckhafer  yeah I think works. 
Sorry about the moving target

rymorgan:
@kieckhafer yep confirmed. We're going to save that feature for a merchanding page functionality in the future

PR #85, opened 2018-05-03
add mobx documentation

kieckhafer:
Start adding documentation for our MOBX stores.

### DIFF
--- a/src/lib/stores/UIStore.js
+++ b/src/lib/stores/UIStore.js
@@ -2,8 +2,25 @@ import { observable, computed, action } from "mobx";
 import getConfig from "next/config";
 
 class UIStore {
+  /**
+   *`query` @observable
+   * @type @observable
+   * @param {boolean} cartOpen - Is the cart drawer open or closed
+   */
   @observable _cartOpen = false;
+
+  /**
+   *`query` @observable
+  * @type @observable
+  * @param {boolean} menuDrawerOpen - Is the menu drawer open or closed
+  */
   @observable _menuDrawerOpen = false;
+
+  /**
+   *`query` @observable
+   * @type @observable
+   * @param {object} appConfig - App config data
+   */
   @observable _appConfig = getConfig();
 
   @computed
--- a/src/lib/stores/UIStore.js
+++ b/src/lib/stores/UIStore.js
@@ -1,6 +1,11 @@
 import { observable, computed, action } from "mobx";
 import getConfig from "next/config";
 
+/**
+ * A mobx store for UI data
+ * @class AuthStore
+ */
+
 class UIStore {
   /**
    *`query` @observable
--- a/src/lib/stores/UIStore.js
+++ b/src/lib/stores/UIStore.js
@@ -8,23 +8,25 @@ import getConfig from "next/config";
 
 class UIStore {
   /**
-   *`query` @observable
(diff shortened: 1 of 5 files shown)


aldeed:
@kieckhafer Some comments.

- The structure, `docs` folder, etc. looks good to me. The docs you have are a great start.
- It should mention that `lib/stores` is where the stores are defined.
- It should have a section explaining how to add a new observable prop to one of the stores.
- I'd rather not see each property documented in this file. Instead each store section could link to that store's code file, and each property can have a doc comment above it in the code (which you can add right away in this PR).

@kieckhafer @mikemurray One question. I compared with one of my past projects that uses mobx, and it doesn't use the getters and setters. So instead of this:

```js
  @observable _token = "";

  @computed get token() {
    return this._token;
  }

  set token(value) {
    this._token = value;
  }
```

It is just

```js
@observable token = "";
```

Can we do that, or was there a certain reason it's different?

kieckhafer:
There is a jsdox-mobx-tags npm package, I'll use this for the comments for now: https://www.npmjs.com/package/jsdoc-mobx-tags

kieckhafer:
@aldeed Since we are using `strict mode` ([enforceActions](https://github.com/mobxjs/mobx/blob/gh-pages/docs/refguide/api.md#configure)), you can't mutate observable props directly, they can only be accessed from action/computed/getter/setter

@mikemurray is that right?

mikemurray:
@aldeed @kieckhafer Yea. Becuase we use `enforceActions` we no longer have direct access to those properties.

mikemurray:
I suppose we don't need the setters/getters, but we would then need to decorate any function that accessed those properties in `@action` at least.

aldeed:
@kieckhafer @mikemurray You can access them anywhere. You just can't set them unless it's in an @action function. So you will need to add `@action` before `fetchAuthToken` in that file for example, but `authState.token` should work fine in a component and will be reactive. I didn't try it, but unless they changed that in the 4.0 release, it should work because I've done it in 3.x

src/lib/stores/UIStore.js
```diff
@@ -1,9 +1,31 @@
...
+ * A mobx store for UI data
+ * @class AuthStore
+ */
+
 class UIStore {
+  /**
+   *`query` @observable
+   * @type @observable
+   * @param {boolean} cartOpen - Is the cart drawer open or closed
+   */
```
aldeed: `@param` is for function parameters and we probably want the type to indicate the actual type rather than `@observable`. Scaler values are actually not special observable values in MobX. It is the parent object that is technically the observable/reactive thing. So really this is just a normal boolean. Maybe format these like so:

```js
/**
 * Is the cart drawer open or closed
 *
 * @type Boolean
 * @default false
 */
```

kieckhafer: 👍 Updated.

There is a mobx jsdoc plugin, I wasn't sure if that wsa comprehensive for what we'd be doing though.

https://www.npmjs.com/package/jsdoc-mobx-tags

aldeed: yeah, does not seem necessary for now

aldeed:
Merged this. I will look into whether we truly need the getters and setters and change those in a different PR if we don't.

PR #90, opened 2018-05-16
chore: fix snyk issues

nnnnat:
Updated `.snyk` to ignore vulnerable deps that can't be updated.


spencern:
I think this is connected to this issue: https://github.com/reactioncommerce/generator-reaction-next/issues/8#issuecomment-389693594

We run snyk against our containerized build in CI.

I'm having success testing like this: 
`docker-compose run web sh -c "cp package.json ../ && cd .. && snyk auth && snyk test"`

PR #94, opened 2018-05-17
feat: add to cart button on PDP

kieckhafer:
_This takes care of the UI portion fo the ticket. The data portion needs some more GraphQL work to be done before it can be done (#95)._

Adds `Add to Cart` button, with quantity chooser, to the PDP.

Uses `uiStore` to read data pertaining to the currently selected variant / option.

## TESTING
- Uncomment the appropriate code in the `handleOnClick` function in `ProductDetailAddToCart.js`
- Navigate to a product page
- Use the quantity chooser to choose a quantity of item
- Press the "add to cart" button
- In your console, see that the correct amount of items was "added/logged"


src/components/ProductDetailAddToCart/ProductDetailAddToCart.js
```diff
@@ -0,0 +1,191 @@
...
+  },
+  incrementButton: {
+    backgroundColor: theme.palette.reaction.activeElementBackground,
+    padding: 6
+  },
+  quantityContainer: {
+    padding: 0,
+    border: `1px solid ${theme.palette.reaction.borderColor}`,
+    backgroundColor: theme.palette.common.white,
+    borderRadius: 2
```
willopez: Lets add a buttonBorderRadius prop to the reaction theme to to provide consistency in future components.

src/components/ProductDetailAddToCart/ProductDetailAddToCart.js
```diff
@@ -0,0 +1,191 @@
...
+    "borderLeft": `1px solid ${theme.palette.reaction.borderColor}`,
+    "borderRight": `1px solid ${theme.palette.reaction.borderColor}`
+  },
+  quantityTypography: {
+    marginBottom: theme.spacing.unit * 2
+  }
+});
+
+
+@withStyles(styles, { withTheme: true })
```
willopez: `{ withTheme: true }` is not necessary is this case, because the theme prop is not being used in the component and `@withStyles` will automatically proved the prop in the styles object.

src/components/ProductDetailAddToCart/ProductDetailAddToCart.js
```diff
@@ -0,0 +1,191 @@
...
+  static propTypes = {
+    classes: PropTypes.object,
+    pdpStore: PropTypes.object,
+    product: PropTypes.object,
+    theme: PropTypes.object
+  };
+
+  static defaultProps = {
+    classes: {},
+    theme: {}
```
willopez: not used in component, therefore unnecessary.

src/components/ProductDetailAddToCart/ProductDetailAddToCart.js
```diff
@@ -0,0 +1,191 @@
...
+    onClick: PropTypes.func.isRequired,
+    option: PropTypes.object,
+    selectedOption: PropTypes.string
+  }
+
+  static propTypes = {
+    classes: PropTypes.object,
+    pdpStore: PropTypes.object,
+    product: PropTypes.object,
+    theme: PropTypes.object
```
willopez: can be removed.

src/components/ProductDetailAddToCart/ProductDetailAddToCart.js
```diff
@@ -0,0 +1,191 @@
...
+                input: quantityInput
+              }
+            }}
+          />
+        </Grid>
+        <Grid item xs={12}>
+          <ButtonBase
+            onClick={this.handleOnClick}
+            className={addToCartButton}
+            fullWidth={true}
```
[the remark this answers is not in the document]
kieckhafer: 👍 I don't think full-width is needed, just forgot to remove.

src/components/ProductDetailAddToCart/ProductDetailAddToCart.js
kieckhafer: 👍

src/lib/stores/PdpStore.js
```diff
@@ -0,0 +1,32 @@
...
+  /**
+   * Selected option on the PDP
+   *
+   * @type String
+   * @default null
+   */
+  @observable _selectedOption = null;
+
+  @computed
+  get selectedOption() {
```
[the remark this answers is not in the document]
willopez: in general, follow mobx patters documented here: https://github.com/reactioncommerce/reaction-next-starterkit/pull/101

src/components/Divider/Divider.js
```diff
@@ -40,17 +40,22 @@ export default class Divider extends Component {
...
     const { classes: { container, item, label } } = this.props;
+
+    if (this.props.label) {
+      return (
+        <div className={container}>
+          <hr className={item} />
+          <Typography component="span" className={label}>{this.props.label}</Typography>
+          <hr className={item} />
+        </div>
+      );
```
aldeed: Generally this is more duplicated identical markup than I'd like to see. A common pattern to avoid this is:

```js
return (
  <div className={container}>
    <hr className={item} />
    {!!label && <Typography component="span" className={label}>{label}</Typography>}
    <hr className={item} />
  </div>
);
```

src/components/Divider/Divider.js
```diff
@@ -40,17 +40,22 @@ export default class Divider extends Component {
     label: PropTypes.string
   }
 
-  static defaultProps = {
-    label: "Available Options"
-  }
-
   render() {
     const { classes: { container, item, label } } = this.props;
```
aldeed: I would still destructure label, and rename the classes one.

```js
const { classes: { container, item, label: labelClass }, label } = this.props;
```

src/components/ProductDetail/ProductDetail.js
```diff
@@ -69,6 +71,7 @@ class ProductDetail extends Component {
               vendor={product.vendor}
             />
             <VariantList product={product} variants={product.variants}/>
+            <ProductDetailAddToCart>Add To Cart</ProductDetailAddToCart>
```
aldeed: I don't see `children` used in the `ProductDetailAddToCart` component, is it seems like you could just do `< ProductDetailAddToCart />` here?

kieckhafer:
@aldeed @willopez `mobx` changes have been pulled in, and my code has been updated accordingly.

src/components/ProductDetailOption/ProductDetailOption.test.js
```diff
@@ -1,18 +1,25 @@
...
     <MuiThemeProvider theme={theme}>
-      <ProductDetailOption
-        onClick={() => true}
-        isSelected={true}
-        option={option}
-      />
+      <Provider uiStore={uiStore}>
+        <ProductDetailOption
+          onClick={() => true}
+          isSelected={true}
```
aldeed: This should be just `isSelected`. The `={true}` is implied.

aldeed:
@kieckhafer There is an issue, which I think is due to data type. In `handleQuantityInputChange`, the value is a string so you get weird behavior. It might be enough to do `{ addToCartQuantity: Number(value) }`, or maybe `{ addToCartQuantity: Math.floor(Number(value)) }` to ensure integer.

may-22-2018 21-36-44

Also, I think the minimum should be 1. Clicking down should not drop it to 0 because why would you ever add 0?

rymorgan:
In terms of the colors -- It's going to be a lot easier/better if we can create our set of variables so that we can just be like @cool-grey or whatever. It seems that the sooner we do that the easier as I also overheard another discussion about color yesterday.  Can we implement these color variables? 

https://zpl.io/aBO3D8A

kieckhafer:
@rymorgan I'll add all these to the stylesheet

src/components/ProductDetailAddToCart/ProductDetailAddToCart.js
```diff
@@ -0,0 +1,204 @@
...
+    classes: {},
+    uiStore: {}
+  };
+
+  state = {
+    addToCartQuantity: 1
+  };
+
+  handleOnClick = () => {
+    // This function currently does nothing. When our GraphQL endpoints are available, we'll use them to add the items to the cart.
```
willopez: @aldeed this looks like a good place to elevate state management logic to a higher level, perhaps to the page level, and remove the mobx dependency for this component. A prop could be called here instead. Thoughts?

src/components/ProductDetailOptionsList/ProductDetailOptionsList.js
```diff
@@ -25,14 +26,17 @@ const styles = (theme) => ({
 });
 
 @withStyles(styles, { withTheme: true })
+@inject("uiStore")
```
willopez: `uiStore` can be removed as it's not being used.

src/components/ProductDetailAddToCart/ProductDetailAddToCart.js
aldeed: Good point. @kieckhafer The parent component, ProductDetail, has uiState, so you can move this code there and you won't need to inject state here.

So at top of the parent render function you would move:

```js
const { uiStore: { pdpSelectedOptionId, pdpSelectedVariantId } } = this.props;

let pdpProductToAddToCart = pdpSelectedVariantId;
if (pdpSelectedOptionId) {
  pdpProductToAddToCart = pdpSelectedOptionId;
}
```

And then do:

```js
<ProductDetailAddToCart variantId={pdpProductToAddToCart} />
```

And change this code here to use the passed in `variantId` prop.

kieckhafer:
@willopez @aldeed updated

PR #100, opened 2018-05-18
Add page size selector

willopez:
Adds a select control to set the the grid's page size.

**NOTE**
I have also updated all components to use material ui v1, I was planning on doing this upgrade on a separate PR, however, I hand already started to update the page selector, and it was easy to update the rest.

1. Add at least 6 products
2. Set default page size to 2 in `UIStore` and update the default `PAGE_SIZES` array in `PageSizeSelector.js` to have values 2, 4 and 6
3. Change the page size by selecting from dropdown and confirm the correct number of items are displayed.

NOTE: You can also test without making any changes if at least 100 products are loaded.


src/components/PageSizeSelector/PageSizeSelector.js
```diff
@@ -0,0 +1,49 @@
...
+@inject("uiStore")
+@observer
+class PageSizeSelector extends Component {
+  static propTypes = {
+    uiStore: PropTypes.object.isRequired
+  }
+
+  handleChange = (event) => {
+    this.props.uiStore.setPageSize(event.target.value);
+  }
```
aldeed: I would make this `handleChange` the same as the one in `Select`, calling `this.props.onChange`. Then call `setPageSize` in `ProductGrid` instead.

My reason is because we may want to copy/import the page size selector elsewhere. Ideally we will move it to our component library for others to use, and they might not use mobx.

So the props for this component should be `onChange` and `pageSize`, and no uiStore prop or injection.

src/components/ProductGrid/ProductGrid.js
```diff
@@ -44,13 +48,26 @@ export default class ProductGrid extends Component {
...
 
+  renderPageSizeSelector() {
+    const { classes } = this.props;
+
+    return (
+      <Grid container spacing={24} className={classes.pageSizeContainer}>
+        <Grid item>
+          <PageSizeSelector />
+        </Grid>
+      </Grid>
```
aldeed: Why not just a simple flexbox div, full width on mobile and right align when bigger than mobile?

src/components/PageSizeSelector/PageSizeSelector.js
willopez: I have elevated event handling to the page level and removed mobx dependencies from `PageSizeSelector`

src/components/ProductGrid/ProductGrid.js
willopez: The grid container and items render to divs with `display:flexbox` and I that spacing can be set with a prop.

willopez:
@aldeed comments addressed.

willopez:
@aldeed comments addressed.

willopez:
@aldeed using only limit param now, and fixed selector selection on refresh.

aldeed:
Found one more bug, where it would lose the correct value in dropdown after clicking refresh and then clicking Next. I pushed a fix for that, so this is now 👍

Issue #103, opened 2018-05-21
build initial Keycloak auth service integration


jshimko:
As part of the initial integration of [Keycloak](https://www.keycloak.org/), start building the oauth2 login flow.  This will be using the new Keycloak implementation started here: https://github.com/reactioncommerce/reaction-keycloak

- [x] create oauth flow login (redirects, etc)
- [x] store token data on client
- [x] inject auth tokens into all GraphQL requests

spencern:
@jshimko when you create the ticket to add docs will you link that here?

PR #114, opened 2018-05-31
Keycloak auth integration

jshimko:
Initial work for #103.

## Details

- add [Keycloak](https://github.com/reactioncommerce/reaction-keycloak) configuration to `next/config`
- add util function to trigger an implicit OAuth flow with Keycloak (link in account dropdown)
- successful auth on Keycloak redirects back to new auth page (`/pages/auth.js`) which then stores the provided auth token in localstorage to be sent to the GraphQL backend via request headers
- rewrote the Apollo Client config to be isomorphic because about 95% of it was duplicated between client and server and several parts (like cookie handling) weren't actually doing anything
- rewrote `withApolloClient` component to use the new isomorphic client config
- updated the `AuthStore` class to allow multiple instances for usage with multiple auth tokens
- added Keycloak and Meteor auth tokens to all headers sent to the backend (when they exist)
- removed all unused cookie parsing/handling

## Test

```sh
# Reaction devserver
git clone git@github.com:reactioncommerce/reaction.git
cd reaction
git checkout release-1.14.0
docker-compose up -d devserver
cd ..

# Keycloak
git clone git@github.com:reactioncommerce/reaction-keycloak.git
cd reaction-keycloak
docker-compose up -d
cd ..

# this repo/branch
git clone git@github.com:reactioncommerce/reaction-next-starterkit.git
cd reaction-next-starterkit
git checkout pr103-keycloak-integration
docker-compose up -d
```

Go to the UI of the starterkit at http://localhost:4000.  Click "login with Keycloak" in the account dropdown at the top right.  You should be redirected to the default Keycloak login page.  Once you login (credentials below), you should be redirected back with your token data.  Your token will then be saved in localStorage and sent along in the headers of all Apollo requests to the GraphQL backend.

Test user...

user: `customer@example.com`
pw: `Password.123!`

For additional info on Keycloak...

https://github.com/reactioncommerce/reaction-keycloak
https://www.keycloak.org/documentation.html


src/components/AccountDropdown/AccountDropdown.js
```diff
@@ -68,9 +79,29 @@ class AccountDropdown extends Component {
...
+      clientId: "reaction-next-starterkit-client",
+      url: "http://localhost:8080/auth"
+    });
+
+    keycloak.init({ flow: "hybrid" }).then((authenticated) => authenticated).catch(() => {});
+
+    keycloak.login({ redirectUri: "http://localhost:4000/auth" })
+      .catch((error) => {
+        throw error;
+      });
```
aldeed: Even though this is MVP, after you get it stable, can you abstract this a bit?

- Add a file src/lib/auth/index.js
- Move the keycloak import and all of this onLogin code there. Do the init stuff immediately in that file.
- Export a generic `login` function from there that wraps `keycloak.login`
- Log an error for the init catch
- Get all of the URL/settings values from new ENV
- Import and call `login` function from `lib/auth` in this file

aldeed: Also, `redirectUri` should include something like `?r=/product/123`, where the value is the full current path + query string. And then AuthPage should use that for the redirect. In other words, you should end up back on the same page you're currently on. This will be important for checkout flow in particular.

src/lib/apollo/initApolloBrowser.js
```diff
@@ -28,6 +29,18 @@ const errorLink = onError(({ graphQLErrors, networkError }) => {
...
+  // get the authentication token from local storage if it exists
+  const token = localStorage.getItem("kc-token");
+  // return the headers to the context so httpLink can read them
+  return {
+    headers: {
+      ...headers,
+      Authorization: token ? `Bearer ${token}` : ""
+    }
+  };
+});
```
aldeed: It might just be that I'm missing something, but in order for SSR pages to be rendered with correct data based on auth, don't you have to use cookies the same way we're doing the `meteor-login-token` auth header? You could just add them to the existing `httpLink` functions in initApolloBrowser and initApolloServer, using `options.authToken` (or maybe use `options.token` and rename the existing to `options.meteorToken`). The handling for the server side session cookie is in the `if (!process.browser)` block in `withData.js`

jshimko:
FYI, this now depends on https://github.com/reactioncommerce/reaction-keycloak/pull/2

src/lib/apollo/initApolloBrowser.js
jshimko: @aldeed So cookies actually didn't have anything to do with the auth token being available on the devserver side.  Every Next.js request originates from the client side and the headers are always being set there, so even in the case of an initial SSR'd request the headers are always present.

The cookies weren't actually doing anything for actual API auth before.  For example, [this token](https://github.com/reactioncommerce/reaction-next-starterkit/blob/master/src/lib/apollo/withApolloClient.js#L52-L61) returned `undefined` 100% of the time regardless of the presence of cookies.

Anyway, I completely rewrote the Apollo configs to be isomorphic and to handle all auth tokens from localstorage (which Keycloak also does) and then pass them up as headers (which the devserver currently expects).  I think localstorage/headers are the more flexible options for several reasons, but the big one is having to whitelist domains in the devserver for any app that wants to send up auth credentials in cookies.  It gets pretty tedious pretty quickly.  Just throwing an auth token in the headers removes that restriction and is a lot more in line with most standalone API's.  Requiring cookies for auth makes the assumption the client will always be a browser.  Since that could easily not be the case at times, I thought headers just felt like the right decision.

But perhaps I'm missing something about why things were done that way in the first place.  Either way, it definitely wasn't doing anything to provide auth to the GraphQL backend, so this seemed like an improvement and a simplification.

jshimko:
@aldeed @spencern I've updated the PR description with details about everything that was changed here, I've made all of the fixed requested in the initial review, and I fixed the failing test.  This should be ready for final review and merging.

aldeed:
@impactmass I committed the cookies / SSR fix. Please update the .env.example after you have created a new keycloak client. Then this can be merged.

package.json
```diff
@@ -35,24 +35,25 @@
...
     "body-parser": "^1.18.2",
     "chalk": "^2.3.2",
     "classnames": "^2.2.5",
+    "cookie-parser": "^1.4.3",
     "express": "^4.16.3",
     "graphql": "^0.13.2",
     "graphql-tag": "^2.8.0",
     "isomorphic-fetch": "^2.2.1",
     "js-cookie": "^2.2.0",
+    "keycloak-js": "4.0.0-beta.1",
```
impactmass: Based on the Keycloak [docs](https://www.keycloak.org/docs/4.0/securing_apps/#_javascript_adapter):

> A best practice is to load the JavaScript adapter directly from Keycloak Server as it will automatically be updated when you upgrade the server. If you copy the adapter to your web application instead, make sure you upgrade the adapter only after you have upgraded the server.

it'll be better if we're dependent on the Keycloak server for this lib. Not sure if that's an easy thing to do here

ticean: @seun I understand why they recommend this on the client side and agree. What about the NodeJS server side? Are we using Keycloak there?

impactmass: A quick peek at the code setup shows that we `require` the keycloak lib only once in this [file](https://github.com/reactioncommerce/reaction-next-starterkit/blob/pr103-keycloak-integration/src/lib/auth/index.js). 

The comment in that file `// trigger oauth login redirect flow with Keycloak (CLIENT SIDE ONLY)` seem to imply that it's only run on the client side.

I'll dbl check to be sure.

aldeed: @impactmass Yes, seems to be client only. So, does that mean we should add a <script> element pointing to some `.js` file path on our own keycloak URL? I'm fine with that approach.

impactmass: @aldeed Yes. I'm doing that right now actually.

PR #116, opened 2018-06-01
Refactor: use new pricing object

willopez:
This PR refactors grid and PDP to use new pricing object.

**Testing**
1. Verify that prices show correctly on grid and PDP.
2. On PDP, ensure that the main display price changes as you select different variants/options.


src/components/ProductDetail/ProductDetail.js
```diff
@@ -42,6 +40,7 @@ class ProductDetail extends Component {
   static propTypes = {
     classes: PropTypes.object,
     product: PropTypes.object,
+    shop: PropTypes.object,
```
mikemurray: Since you're only using `code` as `currencyCode` in the functions `determineProductPrice()` and `render()` why not just pass just that? And also make it required.

---
**Sidenote:**

`shop.currency.code` is quite specific for only needing one thing and`const { shop: { currency: { code: currencyCode } } } = this.props;` Will crash the app if that structure is not exactly defined. I added a ticket #118 about optional chaining.

src/pages/product.js
```diff
@@ -16,13 +16,14 @@ class ProductDetailPage extends Component {
...
-    product: PropTypes.object.isRequired
+    product: PropTypes.object.isRequired,
+    shop: PropTypes.object.isRequired
   }
 
   render() {
     return (
       <Layout>
-        <ProductDetail product={this.props.product} />
+        <ProductDetail product={this.props.product} shop={this.props.shop} />
```
mikemurray: I see In some places you have `currencyCode` as a prop, but for the `ProductDetail` component you're passing in the `shop`. Maybe it would be better to just pass the `currencyCode`?

And if you do pass the currency code alone, then consider making it explicitly required.

```
shop: PropTypes.shape({
  currency: PropTypes.shape({
    code: PropTypes.string.isRequired
  })
})
```

src/components/ProductDetail/ProductDetail.js
```diff
@@ -100,14 +99,38 @@ class ProductDetail extends Component {
     this.selectVariant(variant, option._id);
   };
 
+  determineProductPrice() {
```
mikemurray: Consider adding some documentation

src/components/ProductGrid/ProductGrid.js
```diff
@@ -24,6 +24,7 @@ export default class ProductGrid extends Component {
   static propTypes = {
     catalogItems: PropTypes.arrayOf(PropTypes.object),
     classes: PropTypes.object,
+    currencyCode: PropTypes.string,
```
mikemurray: Should be required if there's no default.

src/components/ProductItem/ProductItem.js
```diff
@@ -17,6 +17,7 @@ import { styles } from "./styles";
 class ProductItem extends Component {
   static propTypes = {
     classes: PropTypes.object,
+    currencyCode: PropTypes.string,
```
mikemurray: Should be required if there's no default.

src/components/VariantList/VariantList.js
```diff
@@ -27,6 +27,7 @@ const styles = (theme) => ({
 export default class VariantList extends Component {
   static propTypes = {
     classes: PropTypes.object,
+    currencyCode: PropTypes.string,
```
mikemurray: Should be required if there's no default.

src/components/VariantItem/VariantItem.js
```diff
@@ -4,13 +4,15 @@ import { withStyles } from "@material-ui/core/styles";
...
 import classNames from "classnames";
+import { priceByCurrencyCode } from "lib/utils";
 
 import styles from "./styles";
 
 @withStyles(styles)
 class VariantItem extends Component {
   static propTypes = {
     classes: PropTypes.object,
+    currencyCode: PropTypes.string,
```
mikemurray: Should be required if there's no default.

src/lib/utils/index.js
```diff
@@ -38,3 +38,31 @@ export function inventoryStatus(product) {
...
+}
+
+/**
+ * Find a variant or option by _id
+ *
+ * @param {Array} variants - A list of variants or options
+ * @param {String} variantId - _id of target variant or option
+ * @returns {Object} - found variant or option or null if none found.
+ */
+export function variantOrOptionById(variants, variantId) {
```
mikemurray: Variants and options are the same things, same schema, so we could just call it `variantById()`. Also, consider using `variant.variantId` below as in the GraphiQL docs it mentioned that `_id` might not always be what you want.

willopez:
@mikemurray Comments addressed.

Issue #56, opened 2018-04-23
Add segment compatible analytics event tracking to the Product Grid


spencern:
We need to start tracking ecommerce analytics events in our starterkit. Initially we'll be tracking all of the Segment V2 Ecommerce Events. You can read the docs for that here: https://segment.com/docs/spec/ecommerce/v2/

Start by add documentation for the events tracked to our event tracking/analytics documentation, explaining the properties tracked and any mapping decisions that were made.

For our current implementation of the Product Grid, we'll need to track two events:

[Product List Viewed](https://segment.com/docs/spec/ecommerce/v2/#product-list-viewed)
[Product Clicked](https://segment.com/docs/spec/ecommerce/v2/#product-clicked)

mikemurray:
@spencern The product price may be a range, should I take the `min` or `max` or `neither`?

spencern:
I'd take the `min` if a range exists

spencern:
maybe create a constant variable for this that we can permit customization of in the .env file later as well?

PR #121, opened 2018-06-05
feat: Add Segment tracking for Product List Viewed and Product Clicked events

mikemurray:
Issue: #56
Impact: **minor**
Type: **feat**

## Issue

Add event tracking for `Product List Viewed` and `Product Clicked` event for the homepage and tag pages.

## Solution

- Add docs outlining data mappings
- Update `lib/tracking/trackProductListViewed.js` to include outlined data mappings
- Add HOC for tracking product clicked event
- Ensure the `Product List Viewed` is triggered when viewing the homepage and tag pages

## Testing

1. Create a segment account and get the API key. Add the key to the `.env` (see `.env.example`)
2. Start the app navigate to the Product grid homepage and tag pages
3. In the segment dashboard under `sources > javascript > debugger` see the event being tracked with proper data mappings

You should see a `Product List Viewed`, and `Product Clicked` events  something like the following examples:

- https://segment.com/docs/spec/ecommerce/v2/#product-list-viewed 
- https://segment.com/docs/spec/ecommerce/v2/#product-clicked

__Note: The mappings are not going to be 100% the same as segments suggestions.__


docs/tracking-events.md
```diff
@@ -250,8 +254,109 @@ class Page extends Component {
...
+
+Data for the Segment e-commerce event [Product List Viewed](https://segment.com/docs/spec/ecommerce/v2/#product-list-viewed), will require values from a `CatalogItemProduct` and `CatalogItemVariant`.
+
+```js
+{
+  // ID of the list being viewed. Should be a tag id or home if it's the homepage
+  list_id: "",
+
+  // Name of tag or Home
+  category: "",
```
kieckhafer: `Product list viewed` does not seem to be getting the `category` data, it's always blank, no matter which list I'm on.

src/lib/tracking/trackProductListViewed.js
```diff
@@ -1,41 +1,18 @@
...
  */
-export default (options) => (Component) => {
-  const component = class trackProductListViewed extends React.Component {
-    render() {
-      return (
-        <Component {...this.props} />
-      );
-    }
+export default (options) => track(({ tag, catalogItems }) => {
+  const products = (Array.isArray(catalogItems) && catalogItems.map((catalogItem) => catalogItem.node.product)) || undefined;
```
kieckhafer: Do we need a protection here to check for `node`, similar to the other Segment PR? We might not, I haven't had any errors yet.

mikemurray: I don't think so, because node should exist if there are items in the array.

mikemurray:
@kieckhafer The category fields just take the first tag from the product. Since we have no real weight mechanic, I figured the first tag must be the most imported, regardless of the page it's on.

So, even a tag page, the category will still be the first tag for that product. The `list_id` is better served to identify what page these related products belong to.

docs/tracking-events.md
mikemurray: @kieckhafer I'm getting the category when I visit a tag page

screen shot 2018-06-12 at 11 01 32 am

mikemurray:
@kieckhafer I'm not sure, but the multiple renders may be related to `withTag` and `withCatalogItems`. The component renders exactly twice, and my assumption is that both are doing a graphql query and that makes the component render twice.

mikemurray:
@kieckhafer added a ticket about the multiple re-renders / multiple tracking events dispatched. https://github.com/reactioncommerce/reaction-next-starterkit/issues/142

kieckhafer:
Looks good with the changes, and the new ticket added to fix the issue.

👍

PR #122, opened 2018-06-05
(feat): 87 nnnnat product grid by tag

nnnnat:
## Solution
Storing a `tag` inside the `routingStore` that can be used in the `catalogItems` query. 
When a nav item is clicked the `routingStore.tag` will update. 
When a `tag` page is loaded the `routingStore.tag` will be set on page load.

## Testing
1. Spin up the storefront with several products and tags loaded in RC.
2. Use the tag nav to see the product grid update based on the current tag page.
3. Hard refresh or go directly to a tag page URL to see the product grid load only the products for the active tag.
4. Update the product grid limit or sortBy, then change pages via the tag nav. The product grid should update the products by tag but retain the limit and sortBy properties.


nnnnat:
@mikemurray updated and ready for review

PR #123, opened 2018-06-06
Refactor _app.js component and move common dependencies to it

willopez:
This PR moves the Layout component to _app.js which makes it persistent across pages.

**Testing**

Verify that transitions between `/, /product/* /tag/*` pages only update the content area, the layout comprised of header, nav and footer should remain persistent.

**NOTE** 
Other enhancements such as adding page transitions will be added on another PR.


willopez:
@mikemurray comments addressed. I moved authentication login to `_app.js` and out of `withTheme`

Issue #109, opened 2018-05-25
Create an "image load" component


kieckhafer:
Create a component which displays a loading screen while an image loads, and then displays the image once it's loading.

Can be based off the image load component in `ProductGridHero.js`

PR #129, opened 2018-06-08
(feat): 109 Image component

nnnnat:
## Issue
#109  
The storefront has a few different ways of displaying images, this PR will consolidate to one Image component. The current image loader is also not working correctly (collapses on itself and never loads a placeholder), this PR will add progressive image loading.

## Solution
Created a few components that render images and replaced all the storefront images with these components.

### Img component
**Progressive Loading**
I've created an `Img` component that does a "Medium/Instagram" like progressive loading effect. To achieve this the component first renders an `img` element with a tiny version of the full resolution image. This low res image should download quickly and will be blurred by the CSS removing the pixelation. The component then creates an `Image` buffer with a `src` of the full resolution image. Once the buffer loads the full resolution image the blurred low res`img` will fade out revealing the full res image.

**Lazy Loading**
The `Img` component takes advantage of the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) to determine if the component is within 50px of the viewport. Once it crosses that threshold it will load the full res image. If the browser doesn't support the `IntersectionObserver` then the component with just do the progressive loading outlined above.

**Responsive**
The `Img` component is responsive-ready with `picture` element support. If provided with an object of image sources the `Img` component will render each source at it's defined breakpoint.

**Component Props**
  - `altText`: The image's alt text
  - `presrc`: The preloading image source string
  - `src`: A single image source string
  - `srcs`: An object of images sources for different screen sizes

### withAssetsPath HOC
To deal with the juggling of asset paths and placeholder images I created a Higher Order Component that uses the `uiStore` to either append the assets path to each image source or replace the source with a placeholder if a source isn't provided.

**Component Props**
  - `presrc`: The preloading image source string
  - `src`: A single image source string
  - `srcs`: An object of images sources for different screen sizes

## Notes
  - **Tag Hero Image:** Because of the `withAssetsPath` HOC adding a full path to a `Tag.heroMediaUrl` in the `Tags` collection you'll end up with a `src` like `"http://localhost:3030https://picsum.photos/1400/600"`. We could maybe make the `withAssetPath` be optional through a prop or `withAssetsPath` could check for an "http" first before adding the assets path. I do see the `Tag.heroMediaUrl` having the same url path as the product media soon so maybe the way this works is fine.
  - **Port to Component Lib:** If we remove the mui `theme` usage and generalized how the `srcs` prop works to take an object like `srcs={{ 500px: "assets/image.medium.jpg", 960px: "assets/image-large.jpg" }}` I think this component could be part of the comp lib.

## Testing
 1. Spin up the Storefront and check out the images.
 2. Test in Chrome/Firefox/Edge to see the lazy loading working.
 3. Test in Safari/IE11 to see the non-lazy loading fallback.


src/components/Img/Img.js
```diff
@@ -0,0 +1,232 @@
...
+
+  /**
+   * Private check for component mount, used in image buffer
+   */
+  _mounted = false;
+
+  /**
+   * Private prop for the img wrapper div, used in intersection observer
+   */
+  _wapper = null;
```
kieckhafer: This should be `_wrapper`

src/components/Img/__mocks__/img.mocks.js
```diff
@@ -0,0 +1,9 @@
+export const mockImg = {
+  URLs: {
+    large: "https://picsum.photos/350/350?image=0",
+    medium: "https://picsum.photos/300/300?image=0",
+    original: "https://picsum.photos/1500/1500?image=0",
+    small: "https://picsum.photos/550/550?image=0",
+    thumbnail: "https://picsum.photos/50/50?image=0"
```
kieckhafer: Should we create some sort of repo / site to use for test images, that we know will always be live (or at least will be out fault if it's not live? Not sure if we should be using a third party to provide testing data like this.

src/components/MediaGallery/__snapshots__/MediaGallery.test.js.snap
```diff
@@ -1,85 +1,3 @@
...
-          <span
-            className="MuiTouchRipple-root-100"
-          />
-        </button>
-      </div>
-    </div>
-  </div>
-</div>
-`;
+exports[`basic snapshot 1`] = `null`;
```
kieckhafer: There is nothing in this snapshot. Is that how it should be?

src/components/MediaGalleryItem/__snapshots__/MediaGalleryItem.test.js.snap
```diff
@@ -1,30 +1,3 @@
...
-    alt=""
-    className="inject-MediaGalleryItem-with-uiStore-img-2"
-    src="/resources/placeholder.gif"
-  />
-  <span
-    className="MuiTouchRipple-root-6"
-  />
-</button>
-`;
+exports[`basic snapshot 1`] = `null`;
```
kieckhafer: There is nothing in this snapshot. Is that how it should be?

src/components/ProductGridHero/__snapshots__/ProductGridHero.test.js.snap
```diff
@@ -14,47 +14,4 @@ exports[`snapshop of tagWithoutMedia 1`] = `
...
-        >
-          <path
-            d="M12,4V2C6.48,2 2,6.48 2,12H4C4,7.58 7.58,4 12,4Z"
-          />
-        </svg>
-      </div>
-    </div>
-  </div>
-</section>
-`;
```
kieckhafer: There is nothing in this snapshot. Is that how it should be?

src/components/ProductItem/__snapshots__/ProductItem.test.js.snap
```diff
@@ -1,94 +1,3 @@
...
-            className="MuiTypography-root-20 MuiTypography-body1-29"
-          >
-            Example Manufacturer
-          </p>
-        </div>
-      </div>
-    </div>
-  </a>
-</div>
-`;
```
kieckhafer: There is nothing in this snapshot. Is that how it should be?

src/components/Img/__mocks__/img.mocks.js
nnnnat: I'm just going to remove these, I only used them for writing the Img component styles.

nnnnat:
@kieckhafer this is ready for another look

kieckhafer:
Lazy loading works when it should, fall back is also working.

Looks good 👍

PR #132, opened 2018-06-11
Add page transitions in progress

willopez:
The current next-router, does not work well with full page transitions. Page transitions cause a full page refresh when clicking on a variant/options. This can be revisited when React Router is integrated with NextJS. Further, there is some discussion in the nextjs queue to add functionality to allow for multiple layout components, this would  also help.


willopez:
Moving to icebox for now.

Issue #159, opened 2018-06-23
Create @withCart decorator to fetch and inject cart data from GraphQL api


spencern:
## Summary
We'll need an `@withCart` decorator that will fetch data from the correct cart and inject it where necessary into the UI. This decorator should use the https://github.com/reactioncommerce/reaction/issues/4354 and https://github.com/reactioncommerce/reaction/issues/4361

Include `load more` sorting and ordering.

aldeed:
Background:
- Anonymous carts are retrieved by ID and token, where both the ID and the token are returned from the `createCart` mutation and both the ID and token are stored locally to the client/session.
- Account carts are retrieved by `accountId` and `shopId`
- When logging in / loading the app, it's possible that there will be both an anonymous cart and an account cart.

Specific things to do:
- Check whether there is an anonymous cart ID and token stored. (Assume these two strings will be in local storage / SSR session, same behavior as the login tokens.)
- If we have an anonymous cart ID + token AND we're logged in, call the `reconcileCart` mutation and use the returned `cart` as the cart object. Assuming the reconciliation is successful, delete the cart ID + token from local storage.
- Else if we are logged in, get the account cart using `accountCartByAccountId` query. (Use primary shop ID for now, though this should probably respect the marketplace shared cart setting eventually.)
- Else if we have an anonymous cart ID + token, get the anonymous cart using `anonymousCartByCartId` query
- Add `cart` prop to the wrapped component
- `Cart.items` is a connection. Request 20 items by default, sorted by newest to oldest `addedAt` date. Add a `loadMoreCartItems` function prop to the wrapped component, which triggers loading of the next 20 if there are more (forward/load-more is the only type of pagination we need).

Trickiness:
- This should be done atomically, i.e., don't repeat the queries for every place where `@withCart` is used.
- In general, all other cart mutations will get back the cart in the response payload and update it in Apollo state, so we don't need to worry about that part here. However, make sure this reconcile/retrieve logic always happens on login and on initial load of the app.
- Finding either type of cart must work on the server for SSR

Issue #128, opened 2018-06-07
MiniCart Component


spencern:
# MiniCart Component
## Overview / Summary
Create a mini-cart component which requires a non-empty items array to be passed in.
This component should display a list of items, a cart summary, and any checkout buttons.
This component is necessary for displaying a mini-cart in a dropdown when an item is added to the cart or when a customer hovers over the cart icon.
This component should be added to the Storefront starter-kit

## Expected use cases
This component will be used whenever a mini version of the cart is needed, such as in the dropdown menu.

### Where should this component live
This component should be created in the starterkit.

## Images
![image](https://user-images.githubusercontent.com/1203639/40945129-94cb00fc-6814-11e8-98d5-83b4113019ab.png)

```jsx
// Pseudo-markup for the MiniCart render function
<div>
  <div scrolling-box>
    <CartItemsComponentProp/>
  </div>
  <div footer>
    <CartSummaryComponentProp/>
    <CartCheckoutButtonsComponentProp/>
    <div cart-fine-print></div>
  </div>
</div>
```

### Props
```js
propTypes: {
  cart: PropTypes.shape(),
  onRemoveItemFromCart: PropTypes.func,
  onChangeCartItemQuantity: PropTypes.func,
  Components: PropTypes.shape({
    CartItemComponent: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
    CartSummaryComponent: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
    CartCheckoutButtons: PropTypes.oneOfType([PropTypes.string, PropTypes.func])
  })
}
```

### Data Fetching
None. A HOC will later be wrapped around this component to do the GraphQL cart query and pass in the cart. For now, mock the incoming `cart` prop in `defaultProps`.

### State Store
All data comes from props. No global state used.

### Routing / Query String
The MiniCart is not directly connected to any route or query string.

### Render Criteria
- The MiniCart rendered when the customer hovers over the cart icon in the navigation header on a desktop browser. (WIRE THIS BEHAVIOR AS PART OF THIS ISSUE)
- The MiniCart is also rendered when the customer successfully adds an item to their cart. When the MiniCart is rendered in this way, the dropdown (another component) should only display for 3 seconds unless the user hovers over the cart icon or the dropdown during that timeout. (FUTURE. DO NOT WIRE THIS BEHAVIOR AS PART OF THIS ISSUE)

### Breakpoints
[...]

PR #169, opened 2018-07-03
feat: Cart Popover component

### LINKED ISSUE #166: Mobile Item Added to Cart Popover Component
# Mobile Item Added to Cart Popover Component
## Overview / Summary
Create a popover component which displays the most recent item added to the cart.
This component should display a single Cart Item and a button with a link to the checkout page.
This component should be added to the Storefront starter-kit
This component should only be displayed on mobile devices

## Expected use cases
This component will be used whenever a mobile user adds a product to their cart.

### Where should this component live
This component should be created in the starterkit.

## Images
image

```jsx
// Pseudo-markup for the mobile cart Popover
<Popover>
  <div name="itemAddedToCart">
    <Image size=thumbnail src=cartItemImage/>
    <span cart-item-qty /> <span cart-item-name /> <span>added to cart</span>
  </div>
  <Button link=checkout">
</Popover>
```

### Props
```js
propTypes: {
  cartItem: PropTypes.shape(),
}
```

### Data Fetching
None. The item should be passed as a prop into this component

### State Store
All data comes from props. No global state used.

### Render Criteria
- The Mobile Item Added to Cart Popover is rendered when a customer in a mobile viewport successfully adds an item to their cart. When the MiniCart is rendered in this way, the dropdown (another component) should only display for 3 seconds unless the user hovers over the cart icon or the dropdown during that timeout. This behavior should be wired up as part of this issue.

### Breakpoints
The Mobile Item Added to Cart Popover should be shown on browsers from 959px and narrower

### Desktop
The Mobile Item Added to Cart Popover is not displayed for desktop users.

#### Normal
Zeplin: https://app.zeplin.io/project/5afc82f5e8d73e250ff9d855/screen/5adf99f52f1dbdd37e7c611c
The normal state of the Popover consists of one section and a button.
The upper section of the popover should display the item that was most recently added to the cart, it's image, qty, and name and the phrase "added to cart"
Below this section should be a single button linking to the checkout

#### Empty
This component should not be rendered without an item to display

kieckhafer:
Resolves #166 
Impact: **minor**
Type: **feature**

## Issue
- Create a popover component which displays the most recent item added to the cart.
- This component should display a single Cart Item and a button with a link to the checkout page.
- This component should be added to the Storefront starter-kit
- This component should only be displayed on mobile devices

## Solution
Create the above mentioned Cart Popover

## Breaking changes
None

## Notes
- This component does not yet have any live data. This should be added in #159. For the time being, there is a `const` with static data for testing purposes. This same data will be passed in via a prop when it's available.
- I don't feel like this looks as good as it can when it's on a screen greater thatn 600px. I think we should re-visit the breakpoints, and use the future modal (no ticket yet) or the mini-cart (#128) to display instead of this component at the 600px breakpoint.

## Testing
1. Open the site on a mobile sized device (less than 960px)
1. Add an item to the cart
1. See that the cart pop-up displays for 3 seconds
1. Add another item to cart
1. Move mouse into display when visible
1. See the pop up doesn't disappear


src/components/ProductDetailAddToCart/ProductDetailAddToCart.js
```diff
@@ -9,8 +9,39 @@ import Typography from "@material-ui/core/Typography";
...
+      label: "Color",
+      value: "Red"
+    },
+    {
+      label: "Season",
+      value: "Summer"
+    }
+  ],
+  currentQuantity: 10,
+  imageUrl: "http://localhost:3000/assets/files/Media/nbYKLrZST5DSF87md/thumbnail/chuttersnap-265339-unsplash.png",
```
nnnnat: Can we make this temp image come from placehold.it? I've been using `http://placehold.it/100` for the comp-library

src/components/CartPopover/CartPopover.js
```diff
@@ -0,0 +1,185 @@
...
+
+@withStyles(styles, { withTheme: true })
+@inject("uiStore")
+@observer
+class CartPopover extends Component {
+  static propTypes = {
+    /**
+     * CartItem data
+     */
+    cartItem: PropTypes.shape({
```
nnnnat: I'm seeing a warning in the browser console about this prop being undefined.
screen shot 2018-07-11 at 8 24 21 am

nnnnat:
I'm also seeing this warning in the console when you click the Add to cart button.
screen shot 2018-07-11 at 8 24 32 am
Not sure if this is being caused by MobX updating state, but you can see in the screenshot that other buttons on the PDP throw a similar warning. I don't think you need to resolve these warnings since it's probably been in the storefront for a bit. I'll create a new issue for it.

PR #170, opened 2018-07-03
Feat add checkout buttons component.

willopez:
Impact: minor
Type:  feature

## Issue
Adds a `CheckoutButtons` component that can be used anywhere checkout buttons need to be rendered. Currently there is only one button, in the future more will be added; when other payment methods are supported. For background details as to why this component was created visit: https://github.com/reactioncommerce/reaction-component-library/pull/110

## Testing
1.  Import component into desired page
2. Verify that checkout button is rendered and it fill the full width of the containing element.
3. Verify clicking on the button takes the user to the `/checkout` URL.


kieckhafer:
LGTM, except for failing Snyk tests. Once those are fixed I think it's good to go.

PR #172, opened 2018-07-10
fix: prevent double render in browser

### LINKED ISSUE #142: Multiple re-renders due to decorators that fetch data using Apollo
## Issue
The decorators, `withTag` `withCatalogItems` specifically on the tag page, cause the component to rerender twice. This isn't a huge issue in itself; however, add tracking and now we get multiple duplicate events triggered in Segment with different data.

First, the data comes in with all of the products for the page without filtering, then the second event has the list with the filtering by tag.

## Question

Should we start considering ways to combine queries for data for into a single query? This would be more in the spirit of GraphQL by making one request for everything we need for a page. Though, it would mean more bespoke queries that may only have a single

## Reproduction

1. Create a segment account and get the API key. Add the key to the .env (see .env.example)
1. Start the app navigate to a tag grid
1. In the segment dashboard under sources > javascript > debugger see the events being tracked
1. If you refresh the tag page, you should see multiple `Product List Viewed` events being fired, once with all products, once with only the products for that tag page (if you have your data set up in a way that has many products in many tags)

nnnnat:
Resolves #142
Impact: **minor**
Type: **bugfix**

## Issue
The storefront would re-render the App twice on certain pages at first visit, ~~this was also causing duplicate tracking events to be sent to Segment~~.

The segment issue wasn't related to the dbl render and is being tracked here #179 

## Solution
We're now only calling `getDataFromTree` method during SSR to prevent the double render in the browser. More info on this solution [here](https://github.com/zeit/next.js/pull/4734)

## Breaking changes
N/A

## Testing
Verify the app no longer fetches data twice on initial load.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,3 @@ See our [full documentation](./docs)
  ## Reference links for development
  ### CSS in JS
  - [Responsive Breakpoints](https://material-ui.com/layout/css-in-js/#responsive-breakpoints)
-
-
-
--- a/README.md
+++ b/README.md
@@ -167,3 +167,6 @@ See our [full documentation](./docs)
  ## Reference links for development
  ### CSS in JS
  - [Responsive Breakpoints](https://material-ui.com/layout/css-in-js/#responsive-breakpoints)
+
+
+
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   },
   "dependencies": {
     "@material-ui/core": "^1.0.0",
-    "@reactioncommerce/components": "^0.2.0",
+    "@reactioncommerce/components": "^0.5.0",
     "@segment/snippet": "^4.3.1",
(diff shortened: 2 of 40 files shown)


nnnnat:
**Status update on this PR**: We've been able to prevent Apollo from fetching the initial data twice and the prevent the app from rerendering on the first load.

There are still outstanding issues with our segment tracking.
 _ The `@track...` decorators are attached to component lifecycle methods like `componentDidUpdate` and it ends up using the `prevProps` instead of the current prop data when sending the event to Segment.
 _  `componentDidUpdate` will sometimes be called without any of the Product data that Segment cares about actually changing. This causes some duplicate events being sent.

**Approach**
Remove the tracking `@` decorators from the component lifecycle methods and explicitly call the tracking event function from inside a lifecycle method only after the product data has changed.

PR #191, opened 2018-07-28
[WIP] Fix server rendering

### LINKED ISSUE #182: SSR does not seem to be working
SSR does not seem to be working. The root request for the home page has a response that does not include the html for the grid. The response has an empty root element: `<div id="__next"></div>` see image below.

empty_response

Further, I see 5 GraphQL requests on initial page load, they `primaryShopId`, `shopQuery`,`tagsQuery` and `catalogItemsQuery`. These should have been executed on the server. See image below.

requests

willopez:
Resolves #182 
Impact: major
Type: fix

## Issue
Server rendering is not working

## Solution
Pass initial state data correctly to client and refactor styles to work correctly.

## Testing
1. Disable javascript in browser
2. Verify pages' HTML is rendered on the server


src/pages/globalStyles.js
```diff
@@ -0,0 +1,9 @@
+export default
```
mikemurray: Should this file live in `lib/theme` since it's not really a page? Unless this is some special next.js thing.

willopez: good point, I'll move to the lib folder.

willopez:
@mikemurray global styles moved to lib directory.

PR #198, opened 2018-08-01
feat: components context

aldeed:
Impact: **minor**
Type: **feature**

## Issue
Passing in `components` everywhere is no fun

## Solution
There is now a components provider wrapped around `App`, which passes down components automatically.

## Breaking changes
None.

## Testing
Verify that the app still works, particularly the cart stuff.


src/pages/_app.js
```diff
@@ -44,26 +46,28 @@ export default class App extends NextApp {
 
     return (
       <Container>
-        <JssProvider
```
aldeed: This looks worse than it is. All I did was put `<ComponentsProvider value={componentsContext}>` around `JssProvider`

PR #219, opened 2018-08-08
fix: add <Link> to allow nav links to be read by SSR

### LINKED ISSUE #196: Top level tags not clickable SSR-only
## Issue

Top level tag navigation items are not clickable when rendering in SSR-only when Javascript is disabled.

## Solution

In `NavigationItemDesktop` and `NavigationItemMobile`, ensure the tags with no sub-tags are wrapped in `<Link>` components.

For tags with children, in SSR-only mode make them clickable to that tag page, but when JS is enabled they should open the sub-menu.

## How to reproduce

1. In Firefox
1. type `about:config` in the address bar and hit enter
1. Search for `javascript.enabled`
1. Double click to disable
1. Launch the starter kit and navigate to the homepage
1. Try clicking on a tag in the tag nav
1. Navigation to that tag page will not happen

kieckhafer:
Resolves #196 
Impact: **minor**
Type: **bugfix**

## Issue
Top level tag navigation items are not clickable when rendering in SSR-only when Javascript is disabled.

## Solution
Wrap top level tags in a `<Link>` component so an `<a>` tag is rendered for SSR / SEO purposes.

## Breaking changes
None.

## Testing
1. In Firefox
1. type `about:config` in the address bar and hit enter
1. Search for `javascript.enabled`
1. Double click to disable
1. Launch the starter kit and navigate to the homepage
1. Inspect a top-level nav element with no children, and see that the nav item is an `<a>` tag

### DIFF
--- a/src/components/NavigationDesktop/NavigationItemDesktop.js
+++ b/src/components/NavigationDesktop/NavigationItemDesktop.js
@@ -36,7 +36,7 @@ const styles = (theme) => ({
     fontSize: "12px"
   },
   primaryNavItem: {
-    textTransform: "capitalize",
+    "textTransform": "capitalize",
     "&:first-child": {
       paddingLeft: "16px"
     }
@@ -144,7 +144,7 @@ class NavigationItemDesktop extends Component {
   }
 
   render() {
-    const { classes: { navigationShopAllLink, primaryNavItem }, navItem } = this.props;
+    const { classes: { primaryNavItem }, navItem } = this.props;
 
     return (
       <Fragment>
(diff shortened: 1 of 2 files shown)


src/components/NavigationDesktop/NavigationItemDesktop.js
```diff
@@ -143,7 +145,7 @@ class NavigationItemDesktop extends Component {
 
     return (
       <Fragment>
-        <Button className={primaryNavItem} color="inherit" onClick={this.onClick}>
+        <Button className={primaryNavItem} color="inherit" onClick={this.onClick} href={`${this.linkPath(navItem)}`}>
```
mikemurray: href={`${this.linkPath(navItem)}`} -> `href={this.linkPath(navItem)}` since it's already a string

kieckhafer: @mikemurray updated

PR #232, opened 2018-08-16
bug: Fix broken mini-cart when no cart is available

### LINKED ISSUE #226: Hovering over cart with no items breaks app
After clearing the client localstorage, I hover over the cart and the app breaks.  The cause it attempting to render cart items when there is no cart yet created. Adding an item to the cart fixes the issue. See image below for more details.

http://g.recordit.co/daZlKbv0Co.gif

### Solution
Check for an existing cart before attempting to render the `MiniCart`, and create a new cart when necessary.

mikemurray:
Resolves #226
Impact: **minor**
Type: **bugfix**

## Issue

When no cart is available, hovering over the cart icon produces an error in the console.

## Solution

Guard against an empty cart and display a cart empty message with actions.

## Breaking changes

none

## Testing

1. With no carts, open the mini cart to see the cart empty message
1. Add an item to the cart, and you should see items in the mini-cart
1. Remove all items from the cart, and you should see the cart empty message in the mini-cart
1. Clicking on "Continue shopping" should take you to the homepage

### DIFF
--- a/src/components/MiniCart/MiniCart.js
+++ b/src/components/MiniCart/MiniCart.js
@@ -22,9 +22,12 @@ const styles = ({ palette, zIndex }) => ({
     backgroundColor: palette.common.white
   },
   emptyCart: {
+    display: "flex",
+    justifyContent: "center",
+    alignItems: "center",
     width: 320,
     height: 320,
-    border: palette.reaction.borderColor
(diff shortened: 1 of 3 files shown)


mikemurray:
@nnnnat fixed broken test. Should be ready for review.

PR #246, opened 2018-08-24
(feat): 225 nnnnat as guest / login checkout

### LINKED ISSUE #225: Create an initial email/login view of the checkout page
# <ComponentName> Component
### Summary description of UI component
This can be either a component or live within the checkout page. The checkout page should render it whenever there is a cart loaded but it does not have an `email` property AND it is an anonymous cart (has no `accountId` prop).

### Rationale for why this component is necessary
This will prompt the shopper for an email address so that the shop can communicate with them about the order. It also serves as a point for encouraging them to create an account.

### Where should this component live
This component should be created in the NextJS Starterkit.

### Implementation Details
The actual implementation will differ a bit from the Zeplin designs. On the left side, there should NOT be username and password fields. Instead there should be two buttons: "Sign in" and "Create a new account". Both buttons should go to the appropriate keycloak page, similar to how the main sign in button is done.

When the "Continue as guest" button is clicked, it should show an error if there is no email address entered. If there is an email entered, it should use the GraphQL mutation to set this email on the cart and update Apollo state. This should then cause the main checkout page to appear.

### Images / Designs of UI component in context

Desktop: https://zpl.io/anKpzlr
Mobile: https://zpl.io/2pOjzQo

### Props

No obvious props, but add any that are needed.

### Data Fetching
The cart will already have been fetched

### State Store
None. All data comes from props.

### Routing / Query String
This is part of the /cart/checkout route page.

## Breakpoints
There are just two variations, desktop and mobile. See Zeplin designs.

## Theme
Unknown

#### Normal
https://zpl.io/anKpzlr

#### Empty
None

#### Loading / retrieving data
None. The cart will have been loaded by the page prior to this.

#### Processing / Waiting for response
The "Continue as guest" button should have `isWaiting` prop while the mutation is happening.

#### Error
Use an ErrorsBlock beneath the email field to show any errors that happen.

#### Active (e.g. button is depressed)
There is no active state

#### Disabled
There is no disabled state

#### Focus
The fields and buttons should accept focus.

#### Normal
https://zpl.io/2pOjzQo

#### Empty
None

#### Loading / retrieving data
None. The cart will have been loaded by the page prior to this.

#### Processing / Waiting for response
The "Continue as guest" button should have `isWaiting` prop while the mutation is happening.

#### Error
Use an ErrorsBlock beneath the email field to show any errors that happen.

#### Active (e.g. button is depressed)
There is no active state

#### Disabled
There is no disabled state

#### Focus
This component cannot be focused. The fields and buttons should accept focus.

## Tests / Requirements
- When I go to the checkout page with an anonymous cart that has items but does not have an `email` property, I see this new view.
- When I click the sign in or create an account button, I'm redirected to keycloak
- When I click the Continue as guest button without entering an address, I should see a required validation error under the field
- When I click the Continue as guest button after entering an address, I should briefly see the spinner on the button and then it should switch to the main checkout view, which should show the email address I entered at the top left.

nnnnat:
Resolves #225 
Impact: **minor**
Type: **feature**

## Solution
  *  Updated the `checkout` page to now show a "login/register/checkout as a guest" view if `cart.account === null || !cart.email`.
  *  Added the `setEmailOnAnonymousCart` mutation and created a handler for the `withCart` HOC.
  *  Added the `GuestForm` component to the login view of the `checkout` page that captures an users email address and sends it to the `setEmailOnAnonymousCart` mutation.
  * Added the `CheckoutEmailAddress` component to the `checkout` page  to display the guest/account email once it's avalible.
 * Updated cart `fragment` GQL query to include `fulfillmentGroups` and `payments` data.
 * Started adding new checkout actions to checkout page.

## Breaking changes
N/A

### Continue as guest checkout.
1. In the storefront, as an anonymous user add a few items to your cart and click the "checkout" button.
2. When the checkout page loads you should see the "login" view with the `GuestForm`. 
4. Enter an email address and click the "Continue as guest" button.
3. The "checkout" view should now display with the newly added email address at the top of the `CheckoutActions`.

### DIFF
--- a/src/componentsContext.js
+++ b/src/componentsContext.js
@@ -6,6 +6,7 @@ import CartItem from "@reactioncommerce/components/CartItem/v1";
 import CartItemDetail from "@reactioncommerce/components/CartItemDetail/v1";
 import CartItems from "@reactioncommerce/components/CartItems/v1";
 import CartSummary from "@reactioncommerce/components/CartSummary/v1";
+import Checkbox from "@reactioncommerce/components/Checkbox/v1";
 import CheckoutAction from "@reactioncommerce/components/CheckoutAction/v1";
 import CheckoutActionComplete from "@reactioncommerce/components/CheckoutActionComplete/v1";
 import CheckoutActionIncomplete from "@reactioncommerce/components/CheckoutActionIncomplete/v1";
@@ -25,7 +26,16 @@ const FontIcon = styled.i`
   vertical-align: middle;
 `;
 
-const iconClear = <svg version="1.1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 14 14" style={{ height: "100%", maxHeight: "100%", verticalAlign: "middle" }}><path d="M9.926 9.105l-2.105-2.105 2.105-2.105-0.82-0.82-2.105 2.105-2.105-2.105-0.82 0.82 2.105 2.105-2.105 2.105 0.82 0.82 2.105-2.105 2.105 2.105zM7 1.176c3.227 0 5.824 2.598 5.824 5.824s-2.598 5.824-5.824 5.824-5.824-2.598-5.824-5.824 2.598-5.824 5.824-5.824z" /></svg>;
+const iconClear = (
+  <svg
+    version="1.1"
+    xmlns="http://www.w3.org/2000/svg"
+    viewBox="0 0 14 14"
+    style={{ height: "100%", maxHeight: "100%", verticalAlign: "middle" }}
+  >
+    <path d="M9.926 9.105l-2.105-2.105 2.105-2.105-0.82-0.82-2.105 2.105-2.105-2.105-0.82 0.82 2.105 2.105-2.105 2.105 0.82 0.82 2.105-2.105 2.105 2.105zM7 1.176c3.227 0 5.824 2.598 5.824 5.824s-2.598 5.824-5.824 5.824-5.824-2.598-5.824-5.824 2.598-5.824 5.824-5.824z" />
+  </svg>
+);
 
 export default {
   AddressForm,
@@ -34,6 +44,7 @@ export default {
   CartItemDetail,
   CartItems,
   CartSummary,
+  Checkbox,
   CheckoutAction,
   CheckoutActionComplete,
   CheckoutActionIncomplete,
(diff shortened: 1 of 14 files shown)


src/componentsContext.js
```diff
@@ -34,20 +47,24 @@ export default {
...
+  Checkbox,
   CheckoutAction,
   CheckoutActionComplete,
   CheckoutActionIncomplete,
   ErrorsBlock,
   Field,
   iconClear,
   iconError: <FontIcon className="fas fa-exclamation-triangle" />,
-  iconValid: (<FontIcon className="far fa-check-circle" />),
+  iconValid: <FontIcon className="far fa-check-circle" />,
```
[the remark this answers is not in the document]
willopez: You'll need to add the `/login` route here; https://github.com/reactioncommerce/reaction-next-starterkit/blob/master/src/pages/_app.js#L75

Further, we could create a new `CheckoutLayout` that can be used for checkout pages such as login, checkout, etc

nnnnat:
@willopez  I believe this is ready for another review. 

### Testing Update
* In step 2 you should now be redirected to the `/login` page and never see the `/cart/checkout` page load.
* In step 4 you should be redirected back to the `/cart/checkout` page

src/pages/login.js
```diff
@@ -0,0 +1,204 @@
...
+    classes: PropTypes.object,
+    setEmailOnAnonymousCart: PropTypes.func,
+    shop: PropTypes.shape({
+      name: PropTypes.string.isRequired,
+      description: PropTypes.string
+    }),
+    theme: PropTypes.object.isRequired
+  };
+
+  static getDerivedStateFromProps({ cart }) {
```
willopez: To prevent a "flash render", caused by  the component's render being called immediately after this method is executed, this login should be move to getInitialProps https://github.com/zeit/next.js/wiki/Redirecting-in-%60getInitialProps%60

src/pages/checkout.js
```diff
@@ -101,28 +106,36 @@ class Checkout extends Component {
     theme: PropTypes.object.isRequired
   };
 
+  static getDerivedStateFromProps({ cart }) {
+    if (cart.account === null && !cart.email) Router.pushRoute("login", "", { customProp: "please next" });
```
willopez: To prevent a "flash render", caused by  the component's render being called immediately after this method is executed, this login should be move to getInitialProps https://github.com/zeit/next.js/wiki/Redirecting-in-%60getInitialProps%60

PR #268, opened 2018-09-06
Checkout complete page

### LINKED ISSUE #244: Checkout Completed Page
# Work
Create a new page in the starter kit, for displaying a new order after checkout. (i.e., "thank you for your order" page).

## Route
`/checkout/order/:id?token=123`

## Details
- Take the ID from the route param.
- If `token` is present in the query string, use that as the order token. Else look in `anonymousOrders` local storage for that ID, to see if it's an anonymous order. If it is, use that order token.
- If we don't have a token and we aren't logged in, show a "Sign In" button with "Sign in to see your order details" above it? @cassytaylor 

## Design
Needed. Possibly just use the same design as the Meteor app for now. @cassytaylor 

Note that unlike the current Meteor app page, we will not allow adding an email on this page because we are not going to allow orders to be created without an email in the first place.

mikemurray:
Resolves #244 
Impact: **minor**
Type: **feature**

## Issue

Add a checkout complete page

## Solution

Added a checkout complete page using many existing components from cart and checkout.

## Breaking changes

none

## Testing

1. In Reaction, sign in, add items to cart and checkout
1. Run through the checkout process and create an order
1. After the order is created, open Robo3T and look for the newly created order
1. Copy the order id and base64 encode it as `reaction/order:<order-id>`
1. In the starter kit, ensure you're signed in as the same user as reaction with the `meteor-login-token`
1. Then navigate to `http://localhost:4000/checkout/order/<order-id>`
1. See the checkout completed page with all relevant data

### DIFF
--- a/src/components/OrderFulfillmentGroup/OrderFulfillmentGroup.js
+++ b/src/components/OrderFulfillmentGroup/OrderFulfillmentGroup.js
@@ -29,7 +29,9 @@ class OrderFulfillmentGroup extends Component {
   static propTypes = {
     classes: PropTypes.object,
     fulfillmentGroup: PropTypes.shape({
-      items: PropTypes.arrayOf(PropTypes.object),
+      items: PropTypes.shape({
+        nodes: PropTypes.arrayOf(PropTypes.object)
+      }),
       data: PropTypes.shape({
         shippingAddress: PropTypes.object
       })
--- a/src/components/OrderFulfillmentGroup/OrderFulfillmentGroup.js
+++ b/src/components/OrderFulfillmentGroup/OrderFulfillmentGroup.js
@@ -13,6 +13,12 @@ const styles = (theme) => ({
   fulfillmentDetails: {
     padding: theme.spacing.unit * 2
   },
+  header: {
+    padding: `${theme.spacing.unit}px ${theme.spacing.unit * 2}px`
+  },
+  headerRightColumn: {
+    textAlign: "right"
+  },
   summary: {
     paddingTop: theme.spacing.unit * 2
   }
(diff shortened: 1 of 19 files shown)


src/pages/checkoutComplete.js
```diff
@@ -0,0 +1,143 @@
...
+          <title>{shop && shop.name} | Checkout</title>
+          <meta name="description" content={shop && shop.description} />
+        </Helmet>
+        <div className={classes.checkoutContentContainer}>
+          <div className={classes.orderDetails}>
+            <section className={classes.section}>
+              <header className={classes.sectionHeader}>
+                <Typography className={classes.title} variant="title">{"Thank you for your order"}</Typography>
+                <Typography variant="body1">{"Your order number is"} <strong>{order && order._id}</strong></Typography>
+                <Typography variant="body1">{"We've send a confirmation email to"} <strong>{order && order.email}</strong></Typography>
```
aldeed: send -> sent

Also the order ID isn't a number technically, so maybe it should say "Your order ID is"?

I think it could also be confusing on a slow connection to see these half-finished sentences for a couple seconds. Maybe wrap these two lines and show only `if (!isLoading && order)`

When loading, it could say "Loading order details..." and/or have a spinner?

cc @cassytaylor

src/routes.js
```diff
@@ -4,6 +4,7 @@ routes
   .add("home", "/", "index")
   .add("cart", "/cart", "cart")
   .add("checkout", "/cart/checkout", "checkout")
+  .add("checkout-complete", "/checkout/order/:orderId", "checkoutComplete")
```
aldeed: Looks like shopProduct is camelCase so change `checkout-complete` to `checkoutComplete`

src/components/OrderFulfillmentGroup/OrderFulfillmentGroup.js
```diff
@@ -0,0 +1,141 @@
...
+          )}
+          {shippingAddress.address1}
+          <br />
+          {(shippingAddress.address2 && shippingAddress.address2 !== "") && (
+            <span>
+              {shippingAddress.address2} <br />
+            </span>
+          )}
+          {shippingAddress.city}, {shippingAddress.region} {shippingAddress.postal} <br />
+          {shippingAddress.country}
```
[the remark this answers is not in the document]
mikemurray: https://github.com/reactioncommerce/reaction-component-library/issues/277

PR #270, opened 2018-09-10
Integrate the FinalReviewCheckoutAction

### LINKED ISSUE #269: Integrate FinalReviewCheckoutAction
Integrate FinalReviewCheckoutAction

willopez:
Resolves #269 
Impact: major
Type: feature

### Summary
Lay the groundwork for integrating the `FinalReviewCheckoutAction` with the starterkit.

TODO: 
- [x] Integrate UI with backend

## Testing
1. Place on order
2. Verify order shows up in reaction
3. Verify cart token and cartId were removed from client
4. Verify corresponding cart was removed from DB


src/components/CheckoutActions/CheckoutActions.js
```diff
@@ -61,13 +65,64 @@ export default class CheckoutActions extends Component {
...
+        fullName: address.fullName,
+        isCommercial: address.isCommercial,
+        phone: address.phone,
+        postal: address.postal,
+        region: address.region
+      };
+
+      const items = cart.items.map((item) => ({
+        addedAt: item.addedAt,
+        price: item.priceWhenAdded.amount,
```
aldeed: This should be `item.price.amount`

src/components/CheckoutActions/CheckoutActions.js
```diff
@@ -61,13 +65,64 @@ export default class CheckoutActions extends Component {
...
+        address1: address.address1,
+        address2: address.address2,
+        city: address.city,
+        country: address.country,
+        fullName: address.fullName,
+        isCommercial: address.isCommercial,
+        phone: address.phone,
+        postal: address.postal,
+        region: address.region
+      };
```
aldeed: You might not need to rebuild the shipping address like this. I needed to in Meteor, but since yours is from GraphQL, I think it's already correct. You should ideally actually do `data` below without caring whether there is even a shipping address on it because in the future "pickup" groups will not have an address.

src/components/CheckoutActions/CheckoutActions.js
```diff
@@ -61,13 +65,64 @@ export default class CheckoutActions extends Component {
...
+
+      const items = cart.items.map((item) => ({
+        addedAt: item.addedAt,
+        price: item.priceWhenAdded.amount,
+        productConfiguration: {
+          productId: item.productConfiguration.productId,
+          productVariantId: item.productConfiguration.productVariantId
+        },
```
aldeed: This could be simplified to `productConfiguration: item.productConfiguration`

src/components/CheckoutActions/CheckoutActions.js
```diff
@@ -61,13 +65,64 @@ export default class CheckoutActions extends Component {
...
+        selectedFulfillmentMethodId: selectedFulfillmentOption.fulfillmentMethod._id,
+        shopId: shop._id,
+        totalPrice: checkout.summary.total.amount,
+        type: group.type
+      };
+    });
+
+    const order = {
+      cartId,
+      currencyCode: "USD", // TODO: Make dynamic once the currency selector is added
```
aldeed: This is not intended to be dynamic. You should take it from `cart.currencyCode`.

src/components/CheckoutActions/CheckoutActions.js
```diff
@@ -91,6 +146,16 @@ export default class CheckoutActions extends Component {
...
     }
 
+    // Order summary
+    const { fulfillmentTotal, itemTotal, taxTotal, total } = summary;
+    const checkoutSummary = {
+      displayShipping: fulfillmentTotal && fulfillmentTotal.displayAmount,
+      displaySubtotal: itemTotal && itemTotal.displayAmount,
+      displayTotal: total && total.displayAmount,
+      displayTax: taxTotal && taxTotal.displayAmount,
+      items
```
aldeed: The `fulfillmentTotal && ` and such are not necessary because they are all required fields in GraphQL schema.

PR #274, opened 2018-09-12
External theme file to control all the component's style.

Akarshit:
Impact: **major**
Type: **refactor**

## Issue
To change the style of a component, that component had to be extended and then replaced because the style for every component was inside that component. 
This is a lot of effort for just changing the style of a single component.

## Solution
This PR extracts the style of all the components to a single file that can be over-ridden by anyone who wants to change the theme of the app.

## Breaking changes
None

## Testing
1. Start the App, no styling should be broken.
1. Change some style in `src/lib/theme/reactionTheme.js`, the style change should be reflected in the UI.


mikemurray:
To add to the above comment, we would need to add `{ name: "<some-name>" }` to every component to make their styles more easily overridable.

We should make sure variables are used where needed and there are no extra, unused variables.

This is also something we should all agree on before any work is attempted, and clarify in documentation the preferred methods customization.

aldeed:
Created an issue for @mikemurray's comment here: https://github.com/reactioncommerce/reaction-next-starterkit/issues/279

Otherwise we will be solving this in other ways per my discussions with @zenweasel so I'm closing this PR.

PR #276, opened 2018-09-12
[WIP] chore: Rename the reaction-api Docker network

focusaurus:
Renames the Docker network on which the GraphQL enabled web services are
attached.

Platform networks in the Docker environment should be named as
*.reaction.localhost. The localhost TLD is reserved and guaranteed to
not conflict with a real TLD.

ref: reactioncommerce/reaction#4447

Resolves #issueNumber
Impact: **major**
Type: **chore**

### Merge Considerations

* The network is only updated in newly created projects.
* To enable network communication, other projects must use this new network name.
* PRs related to reactioncommerce/reaction#4447 should be coordinated.

----

I'm getting my local env set up and preparing to run the tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ _Follow steps as necessary. If you already have Reaction installed, you may be a
 
     You'll need to create a docker network for the GraphQL service and the Reaction Storefront to communicate
     ```
-    docker network create reaction-api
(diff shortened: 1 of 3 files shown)


focusaurus:
Do we really need the user to create the docker network explicitly or can it be handled by our compose.yml entirely?

PR #277, opened 2018-09-12
chore: Rename the reaction-api Docker network

focusaurus:
Renames the Docker network on which the GraphQL enabled web services are
attached.

Platform networks in the Docker environment should be named as
*.reaction.localhost. The localhost TLD is reserved and guaranteed to
not conflict with a real TLD.

ref: reactioncommerce/reaction#4447

Resolves #issueNumber
Impact: major
Type: chore

**Merge Considerations**

- The network is only updated in newly created projects.
- To enable network communication, other projects must use this new network name.
- PRs related to reactioncommerce/reaction#4447 should be coordinated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ _Follow steps as necessary. If you already have Reaction installed, you may be a
 
     You'll need to create a docker network for the GraphQL service and the Reaction Storefront to communicate
     ```
-    docker network create reaction-api
+    docker network create api.reaction.localhost
     ```
     You can run `docker network ls` to verify the network has been created.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -31,7 +31,7 @@ _Follow steps as necessary. If you already have Reaction installed, you may be a
 
     You'll need to create a docker network for the GraphQL service and the Reaction Storefront to communicate
     ```
-    docker network create reaction-api
+    docker network create api
```
aldeed: This would need to be `docker network create api.reaction.localhost`

README.md
```diff
@@ -164,7 +164,7 @@ docker build -t reaction-storefront --build-arg BUILD_ENV=production .
 To start the app in production mode execute:
 
 ```
-docker run -d --name storefront -p ${port}:4000 --env-file .env --network reaction-api reaction-storefront
+docker run -d --name storefront -p ${port}:4000 --env-file .env --network api reaction-storefront
```
aldeed: `api` should be `api.reaction.localhost`

README.md
focusaurus: Hmm. Based on my testing yesterday, it seemed docker wanted to consistently use just a flat name and only for the external bit did we use the FQDN. I'm pretty sure what we use here needs to match the `network:` field in the `docker-compose.yml` files.

aldeed: @ticean ? You had previously mentioned in Slack that the command would be `docker network create api.reaction.localhost`

PR #281, opened 2018-09-14
refactor: 248 nnnnat checkout login

### LINKED ISSUE #248: Move login/guest component to checkout page component
We need to investigate why `getInitialProps` method is not working in the starterkit's Pages. I think it is an implementation issue with our `_document.js` & `_app.js` components but might be an issue with nextjs.

We want to be able to use `getInitialProps` within all our Page components.
[getInitialProps](https://nextjs.org/docs/#fetching-data-and-component-lifecycle)

nnnnat:
Resolves #248 
Impact: **major**
Type: **refactor**

## Issue
When a customer enters checkout without an email or account they'll be redirected to a login page. This happens on the client and can take a few seconds causing a flash of the Checkout before the redirect.

## Solution
 * Updated checkout page component to check if a customer has an account/email in a `componentDidMount` lifecycle method and replace `Router.pushRoute` with `Router.replaceRoute`. Using `.replaceRoute` should preserve the intended back button behavior.
 * Updated login page component to also check if a customer has an account/email in `componentDidMount`.
 * Fixed import paths in a few component tests.
 * Fixed a "missing key" warning in _document `<Head />`

## Breaking changes
N/A

## Testing
1. As an anonymous customer select a product and add it to the cart.
2. Navigate to the `/checkout` page.
    * The page should redirect to the `/login` page.
    * There should be no flash of the "Checkout" view before the redirect happens.
3. Click the browsers back button.
    * The storefront should return to the last page before navigating to the `/checkout` page.


aldeed:
Merging this to get the other changes in. We can do another fix for the redirect if it's still happening.

PR #290, opened 2018-09-17
refactor: nnnnat checkout login

### LINKED ISSUE #248: Move login/guest component to checkout page component

nnnnat:
Resolves #248 
Impact: **minor**
Type: **refactor**

## Issue
When entering checkout as an anonymous customer, you'd sometimes see a flash of "checkout" view before the login/register "entry" view would be displayed. 

## Solution
* Move login/register/guest entry out of the `login` & `checkout` pages and to its own component `Entry`
* Updated `login` & `checkout` pages to use the `Entry` component.
* Created a cart spacific login route ("/cart/login") to keep the curret `login` page working as is.

## Breaking changes
N/A

## Testing
1. As an anonymous customer add a product to your cart and entry checkout.
2. The page should redirect to the login page without showing a flash of the checkout.
3. click the back button to go back to the last page before trying to entry checkout.
4. Enter checkout and fill out the guest form.
5. The page should redirect to "/cart/checkout" and the checkout action view should render.


src/components/Entry/Entry.js
```diff
@@ -0,0 +1,93 @@
...
+const flexWrapper = () => ({
+  alignItems: "stretch",
+  display: "flex",
+  flexDirection: "column",
+  justifyContent: "flex-start"
+});
+
+const styles = (theme) => ({
+  loginWrapper: {
+    ...flexWrapper(theme),
```
mikemurray: The `theme` doesn't have to passed to the mixin since it's not used.

mikemurray:
@nnnnat I also noticed these other two issues

- Checkout login buttons don't seem to route to the auth flow
- Guest checkout seems broken. I add an email and it appears to submit, but nothing happens afterward.

nnnnat:
@mikemurray this should be ready for another review

PR #297, opened 2018-09-18
Updated material-ui, after investigating all updates for changes.

### LINKED ISSUE #291: Update to latest MaterialUI
## Work
Update to latest Material UI packages. Either 3.0.3 or 3.1.0 (just released). Pay attention to breaking changes listed here: https://github.com/mui-org/material-ui/blob/master/CHANGELOG.md

If this causes widespread issues that are difficult to resolve, we may not want to do it for now.

We are hopeful this might resolve #260, too

rosshadden:
Resolves #291 
Impact: **minor**

## Issue
Material UI has updated upstream, but we are on an "old" version (though only a month old, despite being 2 major versions prior).

## Solution
Upgraded the Material UI dependency after thoroughly investigating the versions that have happened since for changes---breaking or otherwise.

## Breaking changes
None.

## Testing
1. Stop/start docker container (`restart` actually didn't work, as it didn't run `npm install`)
2. Test in production build


rosshadden:
@nnnnat Interesting. I had to change a snapshot in the component library, but when I ran the tests for this they all passed. Just ran outside of docker though---maybe that matters.

PR #303, opened 2018-09-18
Cleared before/after params upon tag change

### LINKED ISSUE #262: after/before grid filter should be cleared when filter is changed
## Reproduction
1. Open http://starterkit-staging.devops.reactioncommerce.com/
1. Go to the next page of the grid
1. Select Accessories > Sunglasses > Aviator in the menu.

## Issue
It says there are no results and `after` is still in the URL query string.

## Solution
Remove `before` and `after` filters from the query string whenever we change the tag filter for the grid. It probably makes sense to keep the page count and sorting selections, though.

rosshadden:
Resolves #262 
Impact: **minor**
Type: **bugfix**

## Issue
`before` and `after` query params are persisted across tag changes, which doesn't make sense and breaks expected functionality.

## Solution
Removed `before` and `after` filters from the query string whenever the tag is changed.

## Breaking changes
None.

## Testing
1. Load up `/`.
2. Click `Next` to go to page two.
3. Navigate to a specific tag, like sunglasses
4. The URL before this change had `after` (causing the content shown to not make sense) and after the change has no `after`.


src/pages/tag.js
```diff
@@ -54,6 +54,13 @@ export default class TagShop extends Component {
     if (this.props.catalogItems !== prevProps.catalogItems) {
       this.trackEvent(this.props);
     }
+
+    if (this.props.tag !== prevProps.tag) {
```
willopez: I wonder if this could be moved to `getDerivedStateFromProps` to avoid the initial render before the redirect. Currently, the tag page will start to render and then it redirects to it self without the params, which leads to not the best UX.

rosshadden:
Good idea, changed and tested.

PR #308, opened 2018-09-19
fix: remove global styles

### LINKED ISSUE #292: Links sometimes have default link styling
## Problem
Links sometimes have default link styling (underlined and default link colors) when the page loads/reloads in development. Not sure if this happens on the deployed site.

It likely has something to do with the `globalStyles.js` file not getting loaded or the timing of when it gets loaded.

## Solution
Ideally, remove `globalStyles.js` and set those styles on all Links in a more appropriate and foolproof way.

mikemurray:
Resolves #292 
Impact: **minor**
Type: **bugfix**

## Issue

Default styling for links shows up on occasion.

## Solution

Remove global styles. Links look fine without it?

## Breaking changes

none

## Testing
1. Start app
2. Ensure links across app look as they should

### DIFF
--- a/src/lib/theme/globalStyles.js
+++ b/src/lib/theme/globalStyles.js
@@ -1,9 +1,0 @@
-export default
-<style global jsx>
-  {`
-    a {
-      color: inherit;
-      text-decoration: none;
-    }
-  `}
-</style>;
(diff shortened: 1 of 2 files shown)


aldeed:
@mikemurray It's still happening with globalStyles removed. It happens repeatably for me on the first page load after app startup. (`dc down`, `dc up -d`) As soon as I make a change and it reloads, the problem is gone.

aldeed:
@mikemurray @willopez I pushed a commit on this branch that seems to fix the issue for me (adding styles to the `Link` component). I also updated `styled-components` to latest. I don't think that ended up helping, but it didn't seem to hurt either so we can probably merge that unless you notice any new issues.

PR #320, opened 2018-09-20
feat: Add sign in/up info in auth URL generated by Passport

impactmass:
Part of reactioncommerce/reaction/issues/4640
Type: **feature**

## Issue
When a Starterkit user clicks "Register", we redirect to the Reaction Hydra login page, but it shows the "Sign In" view rather than "Create Account" view.

## Solution
Pass extra info to Hydra about the specific auth flow the user is going through: Sign up OR Sign in. 

## Changes
- Extend Passport configuration to allow passing custom field denoting  "signin" or "signup" flow
- Separate `/auth` route into two distinct `/signin` and `signup` routes to allow requirement above (I couldn't find a way around it with single route)
- Changes needed on the Identity provider to use above update is done in reactioncommerce/reaction/pull/4651.
- Update logout function to delete session in Hydra
- fix: Show logout dropdown onclick Viewer when logged in
- Also moved `splitnames` helper into AuthStore, making it available for reuse beyond Header comp

## Breaking changes
N/A

## Testing
NB: This is to be tested along with reactioncommerce/reaction/pull/4651
1. From Starterkit, click "Sign In"
2. Confirm that you see a "Sign In" form on the Meteor Login page
3. From Starterkit, click "Create Account"
4. Confirm that you see a "Create Account" form on the Meteor Login page

### DIFF
--- a/src/lib/stores/AuthStore.js
+++ b/src/lib/stores/AuthStore.js
@@ -71,9 +71,32 @@ class AuthStore {
     return false;
   }
 
+  // TODO: Temporary workaround until name fields get added from GQL
+  // See https://github.com/reactioncommerce/reaction/issues/4646
+  splitNames(account) {
+    const { name } = account;
+    const firstName =
+      name &&
+      name
+        .split(" ")
+        .slice(0, -1)
+        .join(" ");
+    const lastName =
+      name &&
+      name
+        .split(" ")
+        .slice(-1)
+        .join(" ");
+
+    return {
+      firstName,
+      lastName
+    };
+  }
+
   @action setAccount(account) {
     this.accountId = account._id || null;
-    this.account = account;
+    this.account = { ...this.splitNames(account), ...account };
   }
 }
 
--- a/src/lib/stores/AuthStore.js
+++ b/src/lib/stores/AuthStore.js
@@ -74,8 +74,13 @@ class AuthStore {
   // TODO: Temporary workaround until name fields get added from GQL
   // See https://github.com/reactioncommerce/reaction/issues/4646
   splitNames(account) {
+    let firstName = "";
+    let lastName = "";
     const { name } = account;
-    const [firstName, lastName] = name.split(" ");
+    const nameParts = name && name.split(" ");
+    if (Array.isArray(nameParts)) {
+      [firstName, lastName] = nameParts;
+    }
 
     return {
       firstName,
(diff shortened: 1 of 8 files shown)


mikemurray:
@impactmass It seems that once I've created an account and signed in once. Even if I visit http://localhost:4000/logout to sign out, If I sign in again, I get auto signed in with the last account I signed in with.

1. From the starter kit, create a new account
1. Create a new account
1. On a signed-in starter kit, navigate to http://localhost:4000/logout
1. Sign in / create account again
1. It should auto-sign you into your previous account

impactmass:
@mikemurray thanks, good call 👍 . I'm checking on it. 

This is directly related to having the Hydra sessions working. So when logging out from the Starter kit, there needs to be a call to invalidate the session in Hydra as well. I'll check if it's a 'small' fix that can be added to this PR (or if it needs to be broken out into another).

The reason for Hydra tracking session is here:
https://www.ory.sh/docs/guides/master/hydra/3-overview/1-oauth2#user-login
> The endpoint handler at /login must not remember previous sessions. This task is solved by ORY Hydra. If the REST API call tells you to show the login ui, you must show it. If the REST API tells you to not show the login ui, you must not show it. Again, do not implement any type of session here.

impactmass:
Ready for another review

mikemurray:
@seun **As discussed eariler and just adding it here for reference...** When I try to sign out,  the app never goes to the `logout` route, and if I refresh the window, I don’t see the account dropdown at all.

reaction_and_reaction_ _docker-compose_ _docker-compose_logs_-f_ _279x68_and_slack_-_reaction

What I do see is the following warning in the console:

```
Warning: Did not expect server HTML to contain a <div> in <button>.
```

impactmass:
@mikemurray I've been looking into this after we discussed, and getting different variations of that error:
`Did not expect server HTML to contain a <div> in <button>`, `Expected server HTML to contain a matching <svg> in <span>.` It seems to be an SSR [issue](https://stackoverflow.com/questions/45350360/react-16-warning-warning-js36-warning-did-not-expect-server-html-to-contain-a) that I have yet to figure out.

impactmass:
This commit fixed the issue with the Authstore & re-rendering after logout:
[fix: Update AuthStore to correctly unset account when logged out](https://github.com/reactioncommerce/reaction-next-starterkit/pull/320/commits/6bedc766437830476d952701fbb2f2c7f9b6122d)

nnnnat:
@impactmass I created a merge conflict by approving another PR I'll resolve.

src/lib/stores/AuthStore.js
```diff
@@ -71,9 +71,26 @@ class AuthStore {
     return false;
   }
 
+  // TODO: Temporary workaround until name fields get added from GQL
+  // See https://github.com/reactioncommerce/reaction/issues/4646
+  splitNames(account) {
+    const { name } = account;
+    const [firstName, lastName] = name.split(" ");
```
nnnnat: @mike probably from this? 
Maybe something like `const [firstName, lastName] = name && name.split(" ");`

impactmass: Okay. Will push a fix

PR #342, opened 2018-09-27
Chore add optional chaining

### LINKED ISSUE #118: Optional chaining, and less destructuring
Explore adding [optional chaining](https://github.com/tc39/proposal-optional-chaining) to the babel config

This:
```js
const = { shop: { currency: { code: currencyCode } } } = props
```

Would become this:
```js
const currencyCode = props.shop?.currency?.code || "some default";
```

Destructuring deeply into an object is highly susceptible to failure; as, if any part of that tree is undefined before the final level, there will be an error, and the app will crash.

Even if that value is 100% required, there should be an opportunity to fail gracefully and proving the user options for backing out or trying again.

---

And/or perhaps better guarding of props and not relying on descructuing so much. This would need an update to docs and eslint rules prefering destructuring.

rosshadden:
Resolves #118 
Impact: **minor**
Type: **feature|chore**

## Issue
The [optional chaining proposal](https://github.com/tc39/proposal-optional-chaining) syntax is very convenient, avoiding a lot of checking while drilling down into objects.

## Solution
Added support for optional chaining as a Babel plugin.

## Testing
```javascript
const obj = {
  foo: {
    bar: {
      baz: class {
      },
    },
  },
};

const baz = new obj?.foo?.bar?.baz();
```


mikemurray:
@rosshadden Awesome!

rosshadden:
To be completely honest, I prefer just implementing the current spec. I think adding proposals like this can get messy if they for example get rejected like you mentioned. Stage 3 proposals, which are planned to be added to the spec, are fine with me too. But I would rather not use stages 0, 1, or 2 unless there was a very compelling reason. But then you have edge cases like decorators. They are stage 2, but in this case the reasons are compelling enough.

In terms of this specific feature, which was requested in #118, we need to decide as a team if it is compelling enough. Perhaps in our next Front End meeting.

aldeed:
Added to Oct 10 Frontend Biweekly meeting agenda

aldeed:
Dropping a note that this is now [stage 2](https://github.com/tc39/proposal-optional-chaining/commit/34daace2f8a79af5b0f3b423c8ba09b03cb589b2)

janus-reith:
[Reached Stage 3 now](https://github.com/tc39/proposal-optional-chaining/commit/9948e3db4481562fc4fa0114bbfd42ac9d654013)

PR #349, opened 2018-10-01
Replace passport server-side sessions with cookie session

### LINKED ISSUE #344: Not able to complete authentication on staging deploy
Type: **major**

**Describe the bug**
Our staging deploy of Starterkit at (http://storefront.staging.reactioncommerce.com/) fails to complete the OAuth flow with Hydra instance running at (https://reaction-catalog-sample.staging.reactioncommerce.com/). You either get a Forbidden error message or are not able to login. 

**To Reproduce**
1. Go to http://storefront.staging.reactioncommerce.com/, click Sign In
2. Enter crendentials
3. Note that you are not logged in after entering credentials

**Expected behavior**
Authentication should be complete with a logged in user

impactmass:
Resolves #344
Impact: **major**
Type: **feature+bugfix**

## Issue
Our staging deploy of Starterkit fails to complete the OAuth flow. You either get a Forbidden error message or are not able to login.

This is believed to be caused session management issues. The current Passport setup uses `express-session` which by default stores sessions in memory, which is a problem when the app is deployed on many instances.

## Solution
Make Passport use cookies to store session state (this was our previous setup before the addition of the OAuth flow). So the solution here is to update the Passport configuration to use cookies.

## Testing
1. Test the OAuth flow, and confirm that you can log in and log out.

- If possible, we can/should also deploy this branch and test with the multi-instance server setup.


.env.example
```diff
@@ -13,10 +13,11 @@ KEYCLOAK_REDIRECT_URI=http://localhost:4000/auth
 STRIPE_PUBLIC_API_KEY=ENTER_STRIPE_PUBLIC_KEY_HERE
 OAUTH2_AUTH_URL=http://localhost:4444/oauth2/auth
 OAUTH2_TOKEN_URL=http://hydra:4444/oauth2/token
-HYDRA_ADMIN_URL=http://hydra:4445
```
impactmass: Unrelated to this PR, but discovered this unused env var. It shouldn't be here anymore.

package.json
```diff
@@ -46,8 +46,8 @@
     "chalk": "^2.3.2",
     "classnames": "^2.2.5",
     "cookie-parser": "^1.4.3",
+    "cookie-session": "^2.0.0-beta.3",
```
impactmass: `-beta.3` is their latest version for about a year.

screen shot 2018-10-01 at 1 27 41 pm

I'm not sure if to stay on this version or use a previous. Or maybe consider another package? suggestions welcomed?

impactmass: I also saw some errors in Snyk. I haven't looked well, but it looks like they are from nextjs and not the addition of this package.

PR #352, opened 2018-10-01
Enable Webpack tree shaking

### LINKED ISSUE #351: Enable tree shaking in Webpack
# Enable tree shaking in Webpack

### Summary description
Tree shacking(decode elimination) is currently disabled due to the fact that `modules: "commonjs"` is in the `.babelrc`. This issue is to remove that unnecessary option. See this issue for background: https://github.com/zeit/next.js/issues/4376#issuecomment-409402245

willopez:
Resolves #351 
Impact: minor
Type: feature

## Issue
Wepack tree shacking was not enabled due to the following setting being in our `.babelrc`,
```modules: commonjs```

This PR removes that setting to effectively enable tree shaking(dead code elimination)

## Solution
Follow NextJS' example when using a custom web server: https://github.com/zeit/next.js/blob/canary/examples/custom-server-express/server.js

## Testing
1. Rebuild app
2. Complete the checkout flow, both login and as an anonymous user.
3. Verify there are no errors


rosshadden:
Okay this makes sense. So there are essentially two runtime contexts. One is the startup code (the server) and the other is the Next.js routes and associated code. You had to change everything in the server context, everything they import, and everything they export to use commonjs imports and exports, and then the Next.js context can be ES modules because Next imports them itself through webpack.

Nice work.

spencern:
@willopez @rosshadden it looks like there are two lint errors that are being flagged in CI.

rosshadden:
They might have come from the merge. I'll take a look.

rosshadden:
Yeah my fault. I saw `session` in both sides of the diff and didn't notice `express` changed to `cookie`. Fixed.

PR #362, opened 2018-10-04
Refactor 227 progressive image

### LINKED ISSUE #227: Switch Img to ProgressiveImage
## Work
The `Img` component has been replicated [in the component library](https://stoic-hodgkin-c0179e.netlify.com/#!/ProgressiveImage). Remove `Img` from starter kit and update all `` uses to the one from the component library. For the product grid hero image, the `isHero` prop and its associated CSS padding has been removed, so add a wrapper div with the necessary styles to keep it looking the same.

rosshadden:
Resolves #227 
Impact: **breaking**
Type: **refactor**
Requires: https://github.com/reactioncommerce/reaction-component-library/issues/309

## Issue
We had ported the `Img` component from this project to the component library, but not replaced references here with the new one.

## Solution
Replaced `Img` use with `ProgressiveImage` from the component library.
This is done, but requires the aforementioned component library issue to be closed first. After that point test snapshots might need to be updated.

## Breaking changes
- Removes the `Img` component

## Testing
- TBD


rosshadden:
I believe the CI tests are only failing because of the need for a not-yet-updated component library, since they work locally. I will make it re-run the tests once component library is updated with my required changes.

src/components/ProductGridHero/ProductGridHero.js
```diff
@@ -37,7 +35,7 @@ export default class ProductGridHero extends Component {
       <section className={classes.heroGridContainer}>
         <Grid container spacing={24}>
           <Grid item xs={12}>
-            <Img isHero src={heroMediaUrl} />
+            <ProgressiveImage src={heroMediaUrl} className={classes.heroContainer} />
```
aldeed: Why doesn't a wrapper work here?

```jsx
<div className={classes.heroContainer}>
  <ProgressiveImage src={heroMediaUrl} />
</div>
```

Or even just put the className on `<Grid item`?

I'd prefer to avoid overriding component library styles using className, especially because that mixes styled-components with MUI, but also because people are following these patterns and it's not a pattern we want to encourage.

rosshadden: I tried for quite a while to get something like a wrapper element working (including exactly what you have above). It's because the element that is given the padding by the component library that needs to be overwritten is the `ProgressiveImage` itself, and even with negative padding the desired look is seemingly unachievable without overwriting that property on that element.

aldeed:
@rosshadden I ran this to better understand what you mean. I was forgetting that ProgressiveImage does all of that weird styling stuff and is limited to contain/cover. We should really make a way for that component to render the image with its actual dimensions. But, as a quicker solution for right now, why don't we just use `img` directly? It is simpler and will work properly with any image size then.

```js
const styles = (theme) => ({
  heroImg: {
    width: "100%"
  },
  // ...
});
```

```js
<Grid item xs={12}>
  <img src={heroMediaUrl} className={classes.heroImg} />
</Grid>
```

rosshadden:
@aldeed Okay, done. I agree for this specific case that there is no need to use a component. KISS!

aldeed:
@rosshadden Looks good but there's now a failure from the accessibility linter. I think it's saying you should remove `alt` from the image, but I didn't look into it deeply.

PR #363, opened 2018-10-04
feat: Create hydra client on startup

### LINKED ISSUE #310: Refactor hydra client creation coordination approach
Currently the reaction-hydra process handles the necessary first-time creation of a client via `hydra clients create`. This issue is to discuss separation of responsibilities in terms of which project should handle first-time setup stuff. We talked about moving this into reaction-next-starterkit as it is more of an application concern.

focusaurus:
Resolves #310 
Impact: **minor**
Type: **feature**

## Issue
Move the code to create the hydra client from reaction-hydra to reaction-next-starterkit. This is a task necessary once for each fresh new install (or if you fully destroy all data volumes). It was shipped initially in reaction-hydra but more properly belongs in this repo as its an application concern and reaction-hydra should be providing an agnostic service.

## Solution

Ended up implementing this as a small node.js script as it's a pretty simple HTTP POST with JSON body. I'd be willing to use the golang hydra client CLI program, but I had problems getting that running inside this docker container (posted to their chat for support but no response yet). Certainly willing to keep working to get `hydra` CLI working if others feel that is preferable.

Use our generic `wait-for.sh` script to wait up to 15s for the hydra service to become reachable via TCP.

## Breaking changes

There's new variables in the `.env.example` file. Developers should run `./bin/setup` and the edit their `.env` file to set a unique `OAUTH2_CLIENT_SECRET`, although the defaults are OK for local development. I'm noting this as **minor** impact.

## Testing

1. Test `docker-compose up` for reaction-next-starterkit under the most common scenarios
  1. reaction-hydra is up and available
  1. reaction-hydra is down
  1. reaction-hydra is down but comes up within the 15s timeout
  1. reaction-hydra is up but environment variables point to an incorrect host/port


focusaurus:
Fixing lint errors...

impactmass:
Apologies, I've been unable to get to this PR since I was assigned. I'm pulling it right now to test

bin/create-hydra-client.js
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env node
+
+/* eslint-disable no-console */
+const http = require("http");
+
+/* eslint-disable camelcase */
+const bodyEncoded = JSON.stringify({
+  client_id: process.env.OAUTH2_CLIENT_ID || "reaction-next-starterkit",
+  client_secret: process.env.OAUTH2_CLIENT_SECRET || "CHANGEME",
```
aldeed: Don't do fallbacks in here. This should throw / exit non-zero if any of the necessary env variables are missing.

focusaurus: OK I don't see any validation of env vars in the main app either. Do we have a library we could use for schema checking and just run that against `process.env`? Only thing I could find was `react-props` but that seems to not advise running outside of React.

impactmass:
@focusaurus I'm testing this again now, and getting the same issue as reported yesterday. I'll see if I can narrow the issue down, then post details here

.env.example
```diff
@@ -11,12 +11,15 @@ KEYCLOAK_CLIENT_ID=reaction-next-starterkit
 KEYCLOAK_URL=http://localhost:8080/auth
 KEYCLOAK_REDIRECT_URI=http://localhost:4000/auth
 STRIPE_PUBLIC_API_KEY=ENTER_STRIPE_PUBLIC_KEY_HERE
-OAUTH2_AUTH_URL=http://localhost:4444/oauth2/auth
-OAUTH2_TOKEN_URL=http://hydra:4444/oauth2/token
+OAUTH2_AUTH_URL=http://hydra.auth.reaction.localhost:4444/oauth2/auth
```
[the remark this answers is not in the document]
focusaurus: OK I was able to reproduce this but only after I deleted my existing cookies. With stale cookies I think I was getting errors like this from reaction:

`reaction_1            | 23:01:37.135Z ERROR Reaction: Bearer token is not active`

My next theory is just that the OAUTH2_REDIRECT_URI needs to match the URL you have in your browser address bar. Will dig some more. I suspect all these issues already existed as I don't see how moving this hydra client creation script between containers would have this kind of impact.

focusaurus:
I put `OAUTH2_AUTH_URL` back to how it was and that seems to fix login.

.env.example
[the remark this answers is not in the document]
impactmass: moving the Hydra client creation script between containers did not cause the issue. I think it's just the AUTH_URL host env causing it

.env.example
```diff
@@ -12,11 +12,14 @@ KEYCLOAK_URL=http://localhost:8080/auth
...
 OAUTH2_AUTH_URL=http://localhost:4444/oauth2/auth
-OAUTH2_TOKEN_URL=http://hydra:4444/oauth2/token
+OAUTH2_TOKEN_URL=http://hydra.auth.reaction.localhost:4444/oauth2/token
 OAUTH2_CLIENT_ID=reaction-next-starterkit
 OAUTH2_CLIENT_SECRET=CHANGEME
 OAUTH2_REDIRECT_URL=http://localhost:4000/callback
 OAUTH2_IDP_HOST_URL=http://reaction.api.reaction.localhost:3000/
+OAUTH2_HOST=hydra.auth.reaction.localhost
+OAUTH2_ADMIN_PORT=4445
+PASSPORT_SESSION_SECRET=CHANGEME
```
impactmass: hi @focusaurus, I already approved the PR (it works after your last push). Then I noticed that this PASSPORT_SESSION_SECRET env is not being used anywhere. Is there a reason it's added here? 
Just wanted to confirm that before merging

focusaurus: Looks like that was renamed in a separate piece of work Eric D was doing in 7eef1813f08fecc98b03660e450840af959c8e30. I'll remove it. It is now just called SESSION_SECRET which we have.

PR #394, opened 2018-10-15
feat: Refresh access tokens on 401 errors

### LINKED ISSUE #350: Enable access token refreshing
Type: **major**

## Details

## Notes for implementation

impactmass:
Resolves #350
Type: feature

## Issue
Get a new access token from Hydra.

## Solution
- Setup of a CORS-enabled endpoint in Identity provider that talks to Hydra to retrieve new tokens (reactioncommerce/reaction#4743)
- Update Apollo link setup to catch 401s
- Use endpoint created within IDP in apollo-client to get new tokens
- Write new tokens received to cookie 

## Breaking changes
None

## Testing
1. Start Starterkit on this branch, and Reaction API on this PR reactioncommerce/reaction#4743
2. Run Hydra on master. Set this ENV in the docker-compose yaml before starting the service: ACCESS_TOKEN_LIFESPAN=1m. This will reduce the wait time. The tokens will expire in 60s.
3. Load up a page. Leave it for a while for the token to expire. 
4. Return back to the page, and browser (e.g make a view a product, pause, then come back to add to cart. Try different approaches).
5. Note if there are any token denied errors or 401s. There should be none


impactmass:
Added @mikemurray as primary reviewer.
@aldeed I added you to this PR because of changes around the cookie-session setup (now that it gets changed in the browser when token refreshes). I'd like your comments there.

src/server.js
```diff
@@ -85,6 +86,18 @@ app
...
+      const clientSecret = process.env.OAUTH2_CLIENT_SECRET;
+      const refreshTokenUrl = process.env.OAUTH2_IDP_REFRESH_TOKEN_URL;
+      let encoded = "";
+      try {
+        const str = JSON.stringify({ clientId, clientSecret, refreshTokenUrl });
+        encoded = Buffer.from(str).toString("base64");
+      } catch (error) {
+        logger.error("Error creating encoded config string", error);
+      }
+      res.cookie("oauth_client_config", encoded);
```
aldeed: Why is a cookie needed here? Can you add these three config values to the public app config instead? `publicRuntimeConfig` in `next.config.js`

src/lib/apollo/sessionData.js
```diff
@@ -0,0 +1,75 @@
...
+ * @name createNewSessionData
+ * @method
+ * @private
+ * @param {Object} json session data
+ * @return {String} encoded string containing new session info
+ */
+export function createNewSessionData(json) {
+  let newSessionInfo = "";
+  try {
+    newSessionInfo = Buffer.from(JSON.stringify({
```
aldeed: I don't think most browsers have `Buffer`, but it looks like this is being called from browser code.

src/server.js
```diff
@@ -54,17 +52,20 @@ app
...
+    // The cookie is not `signed` because it's value is to be changed when access_token is
+    // refreshed in the browser.
     server.use(cookieSession({
       // https://www.npmjs.com/package/cookie-session#options
-      keys: [SESSION_SECRET],
       maxAge,
-      name: "storefront-session"
+      name: "storefront-session",
+      httpOnly: false,
+      signed: false
```
aldeed: Removing `httpOnly` does worry me a little, but there may be no other way. Let's document this in plain English and have a couple people review the approach in the docs before merging this. For now, can you just add an `auth.md` file in `docs` in this repo?

Docs should explain
- how cookies are used for auth
- what type of cookies and why
- what is happening to set the cookie during login
- how do the Apollo requests use the cookie/token
- how do we clear the cookie/state at logout
- what happens when token is expired while querying from server code
- what happens when token is expired while querying from client code

src/lib/apollo/sessionData.js
impactmass: Yes, for that I added a browser implementation of Buffer. https://github.com/feross/buffer.

src/server.js
impactmass: Good point. I'll write out docs for it

src/lib/apollo/sessionData.js
aldeed: Oh ok I see the import now

docs/auth.md
```diff
@@ -0,0 +1,28 @@
+# Auth & Cookies
+
```
impactmass: @aldeed @ticean @rosshadden these proposed changes to how we use cookies need a few more eyes to review. I've written these readme to list the changes and will like comments from you all.

rosshadden: Looks good to me @impactmass.

impactmass:
Based on the discussions from this review (especially the case around using not HTTP-only cookie, I'm adding token "refresh flow with redirects only" in #399  (almost finishing up). 
This can serve as a first solution while we figure out the implementation for in browser refresh (new issue here #400). I'll close this PR, but keep the branch up.

PR #399, opened 2018-10-22
feat Renew access tokens on 401 errors using redirects to auth endpoint

### LINKED ISSUE #350: Enable access token refreshing

impactmass:
Resolves #350
Type: feature

## Issue
Renew access tokens on 401 errors by redirecting to re-auth

## Solution
- On 401 errors, redirect to signin endpoint to perform auth and get new access token. Done on both the SSR and browser.

## Breaking changes
None

## Flash of "Not Found"
If going from product grid to a product's detail page and you have an expired token, the app first shows the Not Found component based on this line of code:

https://github.com/reactioncommerce/reaction-next-starterkit/blob/558d94be01491c26c708d8dea98d6e7528a29bdb/src/pages/product.js#L93

The Not Found page is shown while the redirect is happening. Once redirect flow completes, the product component is then displayed properly. I haven't found a way to fix this. 

## Testing
1. Start Starterkit on this branch
2. Run Hydra on master. Set this ENV in the docker-compose yaml before starting the service: ACCESS_TOKEN_LIFESPAN=1m. This will reduce the wait time. The tokens will expire in 60s.
3. Load up a page. Log in (if you are not logged in before). Leave it for a while for the token to expire. 
4. After token expires, perform an action (e.g click to view a product, or add a product to cart).
5. Observe the browser/server console. If a 401 error occurred, confirm that a redirect to /signin starts, and the page is reloaded once the [re]auth is complete.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,6 @@
     "next-routes": "^1.4.2",
     "passport": "^0.4.0",
     "passport-oauth2": "^1.4.0",
-    "passport-oauth2-refresh": "^1.1.0",
     "prop-types": "^15.6.2",
     "react": "16.4.2",
     "react-apollo": "^2.1.3",
--- a/src/containers/catalog/withCatalogItemProduct.js
+++ b/src/containers/catalog/withCatalogItemProduct.js
@@ -22,7 +22,7 @@ export default function withCatalogItemProduct(Component) {
       return (
         <Query query={catalogItemProductQuery} variables={{ slugOrId: query.slugOrId }}>
           {({ data, loading }) => {
-            const { catalogItemProduct } = data;
+            const { catalogItemProduct } = data || {};
             const { product } = catalogItemProduct || {};
 
             return (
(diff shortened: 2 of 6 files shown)


impactmass:
@aldeed see comment in PR description about flash of Not Found. Accessing the PDP doesn't necessarily require a token, but when the app sends an expired token, the API doesn't return data once if finds the token invalid. 

I've tried to pass a value down to the PDP component to use as a check before showing the Not Found page.

ticean:
@aldeed could you take a look at this one when you get a chance? It implements an automatic redirect for a quick, secure solution to token expiry & refresh.

impactmass:
Looks like other tasks have pushed this PR down on @aldeed 's priority list.

But also FYI, I tested this now (after being back from my time off), and I'm discovering that some updates to the Reaction API app has left the OAuth IDP login in a broken state.

I'm writing up an issue for that

PR #401, opened 2018-10-22
fix: integration test finished before resolving

rosshadden:
Impact: **minor**
Type: **bugfix|test**

## Issue
The SSR integration test wasn't actually testing anything, because of how mocha expects async tests.

## Solution
Used mocha's `done` callback workflow.

## Testing

```shell
docker-compose exec web yarn test:integration
```


jeffcorpuz:
Looks good to me! Thank you for the fix.

PR #419, opened 2018-11-07
Enable/Disable Single-page app routing

### LINKED ISSUE #417: Enable/Disable single-page app links
# Feature

Add ENV var + code changes to Enable/Disable single page app (SPA) links.

### Summary description

Add the ability to disable the single page app (SPA) links and 

In some cases, it may be more optimal to load a page as a full page request to allow for better use of caching of the initial HTML payload from the server and reducing the hit to the GQL endpoint for non-user-specific data.

For example, using a least recently used (LRU) cache lib to cache the contents of a page, one could load cached HTML pages in `~20ms`, vs the `~1.2s` it normally takes.

With SPA links enabled, this only affects the initial page load as subsequent initial page loads will incur a penalty of around `300ms` for the GQL request. ( Yes, Apollo has its own internal cache, but this is per browser, per session )

With SPA links disabled, and our theoretical LRU cache enabled, the time to render a static page is reduced to `20ms`. The front-end will then rehydrate the browser app and the user can go about their business.

**These numbers are from testing locally and may not be reflective of all network conditions, or server hardare configurations**

### Expected use cases

- Simple caching for important routes like the grids and PDP pages to improve load times and reduce API overhead
- Puts less load on the GQL endpoint for data that's not changing often
- Better no-JS experience when the browser still be loading and executing the JS app bundle on slow connections

mikemurray:
Resolves #417 
Impact: **minor**
Type: **feature**

## Issue
Description of the issue this PR is solving, why it's happening, and how to reproduce it. This may differ from the original ticket as you now have more information at your disposal.

## Solution

- Add an .env var to enable SPA routing
- Extend next-routes to allow for full page loads
- Update the link component to allow for full page loads
- Update PDP to force it into single-page app mode for variant selection
- Update RoutingStore to for it into single-page app mode when updating the search query (used for the grid filtering)

## Breaking changes

`ENABLE_SPA_LINKS=true` should be placed in your `.env` file

## Testing

With `ENABLE_SPA_LINKS=true`
1. Click through the app and make sure everything works
2. Check the network panel and verify that the page isn't reloaded. (shouldn't see a request for an HTML page)

With `ENABLE_SPA_LINKS=false`
1. Click through the app and make sure everything works
1. Check the network panel and verify that a request for the full page is made
1. Ensure the PDP variant selection does not force a full page load
1. Ensure that the grid filter controls do not force a full page load


impactmass:
@mikemurray thinking about the new ENABLE_SPA_LINKS env; should it have a default value `true`? since that's the previous SPA behaviour. That way if that ENV is not set, it has the default behaviour

mikemurray:
@impactmass then maybe `DISABLE_SPA_LINKS=true` would be better. that way if that doesn't exist, then the default behavior will continue.

mikemurray:
@impactmass I ended up leaving the `.env` var the same name and ensuring that the SPA routing is enabled by default.

I also fixed the prop-type warning you were seeing.

Now on to fixing the broken tests.

mikemurray:
@impactmass ready for review

next.config.js
```diff
@@ -42,7 +42,8 @@ const initExport = {
       url: process.env.KEYCLOAK_URL,
       redirectUri: process.env.KEYCLOAK_REDIRECT_URI
     },
-    stripePublicApiKey: process.env.STRIPE_PUBLIC_API_KEY
+    stripePublicApiKey: process.env.STRIPE_PUBLIC_API_KEY,
+    enableSPARouting: !(process.env.ENABLE_SPA_ROUTING === "false")
```
impactmass: I find this initially confusing. I can figure out what's going on after unpacking each piece.

Could there be a better way of achieving this?

impactmass: I don't know if this is the time we have to bring in this flow that was setup here https://github.com/reactioncommerce/generator-reaction/blob/master/generators/backend/templates/src/config.js

mikemurray:
@impactmass updated the logic and added a comment. Ready for review again.

PR #421, opened 2018-11-07
feat: 366 nnnnat address validation

### LINKED ISSUE #366: Implement address validation.
# Feature Name
Address validation

### Summary description
There should be a UI for address validation where the address returned by the tax/shipping API could be displayed to the user and then s/he can choose the suggested address or continue with the current address.

All the tax calculation and shipping service expect the address to go through their validation APIs before processing the order/request.

### Expected use cases
1. Address validated – the submitted address was correct and / or standardized
2. Address was standardized and is deliverable. Can be updated without informing customer
3. Address was corrected, there are suggested addresses(array of suggested addresses)
4. Address was checked but could not be corrected, suggested addresses have a higher probability of delivery
5. Address could not be validated, address validation service does not support the address country, etc.
6. Unable to contact provider.

The backend for address validation code should be written keeping in mind that other plugin might just want to call their own function for address validation. This would also set up the API which the plugins can follow and the format the plugins have to follow.

nnnnat:
Resolves #366 
Impact: **major**
Type: **feature**

## Issue
#366 

## Solution
Created a `withAddressValidation` Query container that provides Checkout with the `addressValidation` GQL query method as well as any validation results returned from the validation service.

## Breaking changes
N/A

## Testing
*This PR requires Reaction 2.0.0-rc.7*
**Valid addresses**
1. As an Admin within Reaction, open the  "Shop" panel and enable the "Test Validation".
2. As a customer within the Starterkit add an item to the cart and start to check out as a guest.
3. Enter a shipping address that has a `postal` code that starts with "9" (i.e., `90210`).
4. The address should return as valid and checkout should move to the next step.
5. Complete checkout.

**Invalid addresses**
1. As a customer within the Starterkit add an item to the cart and start to check out as a guest.
2. Enter a shipping address that has a `postal` code that **does not** starts with "9" (i.e., `12345`).
3. The address should return as invalid and the AddressReview should display with a InlineAlert explaining the issue.
4. Select the suggested address and submit, checkout should move to the next step.
7. Complete checkout

**Edit an invalid address**
1. As a customer within the Starterkit add an item to the cart and start to check out as a guest.
2. Enter a shipping address that has a `postal` code that **does not** starts with "9" (i.e., `12345`).
3. The address should return as invalid and the AddressReview should display with a InlineAlert explaining the issue.
4. Click "Edit entered address", the UI should switch to the form with the entered address pre-populated.
5. Change the `postal` code to start with a "9" and submit.
6. The address should return as valid and the checkout should move to the next step.
7. Complete checkout

*Note: there is a small UI bug when you try to "Change" a set address, the InlineAlert will sometimes display empty, this is a bug within the component-lib and it being resolved in another PR*

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   },
   "dependencies": {
     "@material-ui/core": "^3.1.0",
-    "@reactioncommerce/components": "0.54.5",
+    "@reactioncommerce/components": "0.55.0",
     "@reactioncommerce/components-context": "^1.0.0",
     "@segment/snippet": "^4.3.1",
     "apollo-cache-inmemory": "^1.1.11",
@@ -54,6 +54,7 @@
     "isomorphic-fetch": "^2.2.1",
     "js-cookie": "^2.2.0",
     "keymirror": "^0.1.1",
+    "lodash.isequal": "^4.5.0",
     "mdi-material-ui": "^5.4.0",
     "mobx": "^4.1.1",
     "mobx-react": "^5.0.0",
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   },
   "dependencies": {
     "@material-ui/core": "^3.1.0",
-    "@reactioncommerce/components": "0.51.2",
+    "@reactioncommerce/components": "0.54.1",
     "@reactioncommerce/components-context": "^1.0.0",
     "@segment/snippet": "^4.3.1",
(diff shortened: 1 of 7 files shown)


machikoyasuda:
1:
screen shot 2018-11-19 at 11 48 30 am

2:
screen shot 2018-11-19 at 11 47 05 am

3:

PR #423, opened 2018-11-08
chore: Add logger messages for 401 error

### LINKED ISSUE #350: Enable access token refreshing

impactmass:
Resolves: NA
Impact: **minor**
Type: **chore**

## Issue
When loading the app after a previous authenticated session and the auth token is expired, the app shows a blank screen. The starterkit server logs shows an error about other failures but not explicit about the 401. 

Looking in the browser network tab can help, as the 401 can be seen there.
The Reaction API service logs shows the `Bearer token is not active` error; but it helps to have it logged here too.

## Solution
Add proper logs that reflect the actual problem (pending when automatic refresh is resolved #350)

## Testing
@Akarshit adding you as reviewer because you ran into this recently. There's really nothing to test other than confirming that the logged message helps to know the actual issue (and prevents getting stuck).

### DIFF
--- a/src/lib/apollo/withApolloClient.js
+++ b/src/lib/apollo/withApolloClient.js
@@ -68,7 +68,12 @@ export default function withApolloClient(WrappedComponent) {
           // Handle them in components via the data.error prop:
           // http://dev.apollodata.com/react/api-queries.html#graphql-query-data-error
           if (error.networkError) {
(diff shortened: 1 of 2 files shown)


aldeed:
@impactmass Needs rebase to resolve conflicts now

PR #453, opened 2018-12-07
feat(tlp metadata): render metafield as metadata on TLP

### LINKED ISSUE #444: Tags: An admin user's metadata appears on the Tag Listing Page
# Tag metadata

### Summary description

As a Reaction 2.0 Admin user, I can add metadata (keywords, description, Facebook graph) for a specific tag, to be displayed on its Tag Listing Page.

The metadata keywords and description added by admin in https://github.com/reactioncommerce/reaction/issues/4846

Tag pages need unique metadata to improve their SEO, trackability, and shareability.

### Expected use cases

- Improved SEO
- Improved trackability with segment/analytics providers
- Improved sharability on social networks

### Expected work

Add meta tags to the document head of the tag pages based on the metadata available for that tag. 

**The fields should include:**
- Keywords
- Description
- Facebook OpenGraph
  - Title `og:title`
  - URL `og:url`
  - Description `og:description`
  - Image `og:image`
  - Facebook App ID `fb:app_id`  **optional**
  - Localization `og:locale`

### Resources 
- Facebook sharing guide: https://developers.facebook.com/docs/sharing/webmasters#markup

machikoyasuda:
Resolves #444 
Impact: **minor**
Type: **feature|docs**

## Issue
- A Tag Listing Page (TLP) needs metadata in the `<head>`, like `keywords`, `description` and OpenGraph tags, like `og:image`.

## Solution
- Back-end: Use a Tag's existing `metafields` field to store data to be used in the HTML document's `<head>`. Specifically, use `namespace` as `"metatag"`, to distinguish metatags. Then, use `scope`, `key` and `value` fields for the various values of the metatag.
- Front-end: Use `scope`, `key`, `value` fields to construct metadata HTML tags like this:

```js
metafield = [
  scope: "property",
  key: "og:type",
  value: "product",
  namespace: "metatag"
]
```

```html
<meta property="og:type" content="product" />
```
Regular test:
screen shot 2018-12-07 at 12 33 36 pm

Test with different description sources:
screen shot 2018-12-12 at 3 57 25 pm
screen shot 2018-12-12 at 3 54 57 pm

## Breaking changes
None

If a Tag has no metafields, default metatags should appear: `{ name: "description", content: shop && shop.description }`

screen shot 2018-12-07 at 12 33 18 pm

## Testing
1. Test a tag with metafields
2. Test a tag without metafields
3. Test a tag with non-valid metafield data
4. Test the case of a tag with `description` metafields:

Test data:
```json
    "metafields" : [ 
        {
            "scope" : "property",
            "key" : "og:type",
            "value" : "product",
            "namespace" : "metatag"
        }, 
        {
            "scope" : "property",
            "key" : "og:title",
            "value" : "EMS | Shirts",
            "namespace" : "metatag"
        },
        {
            "scope": "name",
            "key": "description",
            "value": "this is my tag description from metafields",
            "namespace": "metatag"
        }
    ]
```
## Post-merge
Add metatag data to some Tags to sample data for testing


machikoyasuda:
@mikemurray I'm getting this console warning.. not sure how to fix it:

screen shot 2018-12-07 at 12 06 59 pm

mikemurray:
The error is in the `src/containers/catalog/withCatalogItems.js`  HOC. The `tag` prop appears to be used as an object, but the prop is defined as a string. The prop-type definition is likely wrong.

src/pages/tag.js
```diff
@@ -74,6 +74,20 @@ export default class TagGridPage extends Component {
     this.props.uiStore.setSortBy(sortBy);
   };
 
+  renderHeaderMetadata = (metafields) => {
+    const metadata = [];
+    metafields.forEach((field) => {
+      if (field.valueType && field.valueType === "header") {
```
mikemurray: `namespace` seems like a better fit for types like `header`, `custom` etc, where `valueType` would be better for `string`, `boolean`, `number` in cases where the meta value shouldn't be treated as a string.

Also, `header` should be `head` or `metatag` as to not be confused with the `<header>` tag, or a block on the page that's considered the header.

src/pages/tag.js
```diff
@@ -92,7 +106,12 @@ export default class TagGridPage extends Component {
       <Fragment>
         <Helmet
           title={`${tag && tag.name} | ${shop && shop.name}`}
-          meta={[{ name: "description", content: shop && shop.description }]}
+          meta={
+            tag.metafields && tag.metafields.length > 0 ?
+              this.renderHeaderMetadata(tag.metafields)
+              :
+              [{ name: "description", content: shop && shop.description }]
```
mikemurray: If I define metafields for a tag, but omit the `description` field, then I don't get the shop description at all.

We should try to keep the shop's description if one isn't defined for the tag. React Helmet will de-dupe fields, but I'm not sure if it works in the same array; though it's worth trying either way.

src/pages/tag.js
```diff
@@ -74,6 +74,20 @@ export default class TagGridPage extends Component {
...
   };
 
+  renderHeaderMetadata = (metafields) => {
+    const metadata = [];
+    metafields.forEach((field) => {
+      if (field.valueType && field.valueType === "header") {
+        const metadatum = {};
+        const { scope } = field;
+        metadatum[scope] = field.key;
+        metadatum.content = field.value;
```
mikemurray: Since `scope` is de-structured, then perhaps the other properties like `key`, `value`, and maybe even `valueType` / `namespace` should also be de-structured.

machikoyasuda:
Ready for re-review! @mikemurray

src/pages/tag.js
```diff
@@ -98,12 +112,18 @@ export default class TagGridPage extends Component {
...
+    const description = [{ name: "description", content: shop && shop.description }];
 
     return (
       <Fragment>
         <Helmet
           title={`${tag && tag.name} | ${shop && shop.name}`}
-          meta={[{ name: "description", content: shop && shop.description }]}
+          meta={
+            tag.metafields && tag.metafields.length > 0 ?
+              this.renderHeaderMetatags(tag.metafields).concat(description)
```
mikemurray: If I add the following meta field to a tag, I get a duplicated description metatag with the shop description.

```js
        {
            "scope" : "name",
            "key" : "description",
            "value" : "THE OVERRIDEN DESCRIPTION",
            "namespace" : "metatag"
        }
```

machikoyasuda: @mikemurray I updated the `renderHeaderMetatags` method to check for whether a `description` has been added, and if it has not, to add the `this.props.shop.description` there.

PR #455, opened 2018-12-10
fix: Add proper isMounted checking

### LINKED ISSUE #451: Add proper isMounted checking
## Work
1. Search for `this.setState` and `this.forceUpdate`.
2. For any that are in a callback/then, or after an await, add isMounted check as described [here](https://docs.reactioncommerce.com/docs/react-best-practices#setting-state-asynchronously)

For background, see https://github.com/reactioncommerce/reaction-next-starterkit/issues/194

impactmass:
Resolves #451 
Impact: **minor**
Type: **fix|enhancement**

## Issue
Trying to set state asynchronously, sometimes you may run into a warning stating that you can't call `setState`, or `forceUpdate` on a non-mounted component.

## Solution
1. Search for `this.setState` and `this.forceUpdate`.
2. For any that are in a callback/then, or after an await, add isMounted check as described [here](https://docs.reactioncommerce.com/docs/react-best-practices#setting-state-asynchronously)

## Breaking changes
N/A

## Testing
- This is more of a code review and test to confirm that no regression is created for the components touched

### DIFF
--- a/src/components/CheckoutActions/CheckoutActions.js
+++ b/src/components/CheckoutActions/CheckoutActions.js
@@ -67,6 +67,7 @@ export default class CheckoutActions extends Component {
   };
 
   componentDidMount() {
+    this._isMounted = true;
     const { cart } = this.props;
     // Track start of checkout process
     this.trackCheckoutStarted({ cart, action: CHECKOUT_STARTED });
@@ -90,6 +91,10 @@ export default class CheckoutActions extends Component {
     ) {
       this.handleValidationErrors();
     }
+  }
+
+  componentWillUnmount() {
+    this._isMounted = false;
   }
 
   @trackCheckoutStep()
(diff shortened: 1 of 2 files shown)


impactmass:
@aldeed I'm opening a PR for #451 here, and I'm not sure what the tests instructions should be, since the warnings that led to creating the issue stopped showing up according to [this comment](https://github.com/reactioncommerce/reaction-next-starterkit/issues/176#issuecomment-422949129)

I wanted to confirm if there's some other testing other than code review, before I attach a reviewer to look at the code.

aldeed:
@impactmass I think a code review is sufficient. You may be able to get it to happen by quickly routing to the checkout page, then back to another page, and back and forth a bunch.

aldeed:
@mikemurray We discussed this in the frontend meeting awhile back. Might have been while you were out. But the consensus was to stick with isMounted until someone standardizes how cancellable promises work. They are a lot more overhead to implement, so it would suck to have to rewrite everything if standardized cancellable promises do not end up matching how we do it.

In the [docs](https://docs.reactioncommerce.com/docs/react-best-practices#setting-state-asynchronously) that I wrote after that decision, I added an explanation of our reasoning:

> You may hear it said that canceling promises on unmount is a better approach, and that's probably true, but there is not yet an established pattern for this in wide use so we've decided to use isMounted until the React maintainers have a concrete recommendation.

mikemurray:
@aldeed Is there an issue somewhere tracking this as something we should go revisit later one a standard comes out?

aldeed:
@mikemurray There is [now](https://github.com/reactioncommerce/reaction-next-starterkit/issues/460) 😸

PR #461, opened 2018-12-14
Improve loading pattern for tags and navItems

### LINKED ISSUE #458: Tags loading slowly
Type: **minor**

**Describe the bug**
When the number of Tags is more (> 100). The tags start loading very slowly. And also they come one by one.

**To Reproduce**
1. Add a lot of Tags.
2. Load the homepage.
3. Observe that the page loads first. Then the tags start appearing one by one.

aldeed:
Resolves #458
Impact: **minor**
Type: **performance**

## Issue
For a shop with >200 tags, paging would be necessary when loading tags to build the navigation tree, and this was causing slow loading of the top nav and/or errors.

## Solution
The way we were doing the loading was susceptible to extra looping due to calling `fetchMore` within a `render` function. However, Apollo's `Query` component does not give any better ways to do it. There is an `onCompleted` function, but it seems to get called only sporadically (maybe only when loading from network and not from cache).

Furthermore, our aim with loading all of the tags is actually to build a single tree for nav, so it's beneficial if we can loop through and load all of the pages, and then build the nav, prior to loading the app at all.

So for now I am doing all of the loading and building in the `getInitialProps` of the `_app.js` layout. I'm then passing those to a MobX Provider, which components that need tags or navItems are now injecting.

This seems to work much better. It should be loading from the Apollo cache after the first time, so it should be good performance as far as API and database hits.

We will soon rewrite this to load the entire nav tree at once, using the new `defaultNavigationTree` query, but that isn't released yet. At that time, we may be able to revert back to using a Query component since it will then be a single query rather than looping pages.

## Breaking changes
If you have custom components using `tags` or `navItems`, you'll need to inject those props from MobX and the tag structure has been flattened a bit.

## Testing
1. Load approx 250 tags into the database for your primary shop. Make sure they are all visible and not deleted, and either top-level or listed as a subTag for another tag.
2. Test storefront home page, tag page, and product detail page.
    - Verify that the top navigation looks as expected and works correctly on all pages
    - Verify that the breadcrumbs on all pages look right and work correctly.

Nav loading should be relatively fast and without errors.


aldeed:
@Akarshit Unfortunately I could not find any easy fix other than a significant rewrite of how tags and nav are loaded. If you are able to pull this in to the fork fixing conflicts, it would be good to verify that this fixes your issue.

Akarshit:
Yes it works. The tags load up instantaneously.

PR #463, opened 2018-12-17
Obey url rewrites in browser-side routing

### LINKED ISSUE #452: Tags: URL rewrites for tag pages are overridden when query params change
Type: **minor**

**Describe the bug**
Using a proxy like nginx or skipper to do rewrites of URL's for the tag pages do not stick. They are overridden and revert back to the system defined route when query params change.

**To Reproduce**
1. Setup the full Reaction Platform
1. Add rewrite rewrite rule for from: `/tags/example/product` to: `/tags/example-product`
1. Navigate to `http://localhost:<proxy-port>/tags/example/product`
1. See that you do get the correct page
1. After a second or two, see that the URL has changed to `http://localhost:<proxy-port>/tags/example-product?limit=20`

**Expected behavior**

URL should remain in the same style as the rewrite.

`http://localhost:<proxy-port>/example/product?limit=20`

mikemurray:
Resolves #452 
Impact: **minor**
Type: **feature**

## Issue

Rewrite rules by nature mask the actual URL the underlying application uses for routing. This is problematic when dealing with single-page app routing as it has no idea that route exists, resulting in 404's.

## Solution

Update starter-kit to accept a `Request-Path` from the proxy handling the rewrite rules in order to make a new route entry in the browser-side app.

## Breaking changes

none.

## Testing
1. With the reaction platform, and steps outlined in other repos
1. Start reaction platform
1. In the Redirect URLs admin panel, add a new rewrite rule for a tag page
1. For example from: `/example/route` to: `/tag/shop`
1. Navigate to `localhost:9090/example/route`
1. See the tag page you expect
1. Adjust the page limit/price filters
1. See that the URL in the address bar remains `/example/route`


ross:
> @ross You are the "does this make sense?" reviewer.

Looks like you want @rosshadden here.

mikemurray:
Thanks @ross!

PR #471, opened 2019-01-04
fix: fix logic for retrying 401s with silent re-auth

### LINKED ISSUE #470: Handle 401s better
@griggheo got an error like this:

```
[Network error]: SyntaxError: Unexpected token U in JSON at position 0
{ Error: Network error: Unexpected token U in JSON at position 0
    at new ApolloError (/usr/local/src/node_modules/apollo-client/bundle.umd.js:121:32)
    at /usr/local/src/node_modules/apollo-client/bundle.umd.js:1188:45
    at /usr/local/src/node_modules/apollo-client/bundle.umd.js:1619:21
    at Array.forEach (<anonymous>)
    at /usr/local/src/node_modules/apollo-client/bundle.umd.js:1618:22
    at Map.forEach (<anonymous>)
    at QueryManager.broadcastQueries (/usr/local/src/node_modules/apollo-client/bundle.umd.js:1613:26)
    at /usr/local/src/node_modules/apollo-client/bundle.umd.js:1115:35
    at process._tickCallback (internal/process/next_tick.js:68:7)
  graphQLErrors: [],
  networkError:
   { SyntaxError: Unexpected token U in JSON at position 0
       at JSON.parse (<anonymous>)
       at /usr/local/src/node_modules/apollo-link-http-common/lib/bundle.umd.js:45:29
       at process._tickCallback (internal/process/next_tick.js:68:7)
     response:
      Body {
        url: 'https://rc-staging.bobs-test.io/graphql-alpha',
        status: 401,
        statusText: 'Unauthorized',
        headers: [Headers],
        ok: false,
        body: [PassThrough],
        bodyUsed: true,
        size: 0,
        timeout: 0,
        _raw: [Array],
        _abort: false,
        _bytes: 12 },
     statusCode: 401,
     bodyText: 'Unauthorized' },
  message: 'Network error: Unexpected token U in JSON at position 0',
  extraInfo: undefined }
[Network error]: SyntaxError: Unexpected token U in JSON at position 0
(node:29) UnhandledPromiseRejectionWarning: Error: Network error: Unexpected token U in JSON at position 0
    at new ApolloError (/usr/local/src/node_modules/apollo-client/bundle.umd.js:121:32)
    at /usr/local/src/node_modules/apollo-client/bundle.umd.js:1188:45
    at /usr/local/src/node_modules/apollo-client/bundle.umd.js:1619:21
    at Array.forEach (<anonymous>)
    at /usr/local/src/node_modules/apollo-client/bundle.umd.js:1618:22
    at Map.forEach (<anonymous>)
    at QueryManager.broadcastQueries (/usr/local/src/node_modules/apollo-client/bundle.umd.js:1613:26)
    at /usr/local/src/node_modules/apollo-client/bundle.umd.js:1115:35
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:29) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 3)
(node:29) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

This appears to be a 401 but for some reason isn't handled well.
- Try to replicate and improve. It might be due to resetting the database while logged in?
- If we're unauthenticated and we aren't able to refresh (such as because the account has been deleted), we should be clearing the login cookie so that it just looks as though you've been logged out.

aldeed:
Resolves #470 
Impact: **minor**
Type: **bugfix**

## Issue
Due to GraphQL server update to Apollo Server 2.0 as well as subsequent changes in starterkit, the earlier work to silently re-authenticate when a 401 error is detected stopped working.

## Solution
Revamped how we're doing it to be more foolproof.

## Breaking changes
None

## Testing
I recommend testing against https://github.com/reactioncommerce/reaction/pull/4894 or the re-auth from browser will not work.

There are two ways this token refreshing is supposed to work, both of which need to be tested. The setup is the same for both:
- Add `- ACCESS_TOKEN_LIFESPAN=1m` in the `environment` section of reaction-hydra project's `docker-compose.yml` file.
- `dc down` and `dc up -d` reaction-hydra
- Log out of starterkit, and then log back in.

This will force auth tokens to expire after only 1 minute, making it faster to test this.

### Test silent re-auth in browser
1. After waiting at least a minute since you logged in, click something in the app to cause a GraphQL request to happen. Make sure it's something you haven't recently clicked or Apollo will actually be using its cache. If you don't see a `graphql-alpha` request in the browser network log, then keep clicking until you do.
2. You should see a 401 response and "Attempting silent re-auth" message in the browser console, which may immediately disappear. The page may flash and then you should still be logged in, or worst case it should bring you to the login page if it was not able to silently refresh the token.

### Test silent re-auth from NextJS server
1. After waiting at least a minute since you logged in (or since you did the last token refresh test), refresh the page.
2. You should see an "Attempting silent re-auth" message in the NextJS server log. When the page finishes refreshing, you should still be logged in.

### DIFF
--- a/src/components/VariantItem/VariantItem.js
+++ b/src/components/VariantItem/VariantItem.js
@@ -63,10 +63,16 @@ class VariantItem extends Component {
 
     const variantPrice = priceByCurrencyCode(currencyCode, pricing);
 
+    const className = classNames(variantButton, {
+      [activeVariant]: isActive
+    }, {
+      [soldOutVariant]: variantInventoryStatus && variantInventoryStatus.type === "SOLD_OUT"
+    });
+
     return (
       <ButtonBase
         disableRipple
-        className={classNames(variantButton, { [activeVariant]: isActive }, { [soldOutVariant]: variantInventoryStatus && variantInventoryStatus.type === "SOLD_OUT" })}
+        className={className}
         onClick={this.onClick}
       >
         <Typography component="span" variant="body1">
--- a/src/lib/apollo/initApollo.js
+++ b/src/lib/apollo/initApollo.js
@@ -7,7 +7,12 @@ import { onError } from "apollo-link-error";
 import { InMemoryCache } from "apollo-cache-inmemory";
 import fetch from "isomorphic-fetch";
 import getConfig from "next/config";
+import logger from "../logger";
 import { omitTypenameLink } from "./omitVariableTypenameLink";
+
+const SIGN_IN_PATH = "/signin";
+const STATUS_FOUND = 302;
+const STATUS_UNAUTHORIZED = 401;
 
 // Config
 let graphqlUrl;
@@ -41,14 +46,26 @@ const create = (initialState, options) => {
 
     if (networkError) {
       const errorCode = networkError.response && networkError.response.status;
-      // In browser, if a 401 Unauthorized error occurred, redirect to /signin.
-      // This will re-authenticate the user without showing a login page and a new token is issued.
-      if (errorCode === 401) {
-        if (process && process.browser) Router.pushRoute("/signin");
-      }
+      if (errorCode === STATUS_UNAUTHORIZED) {
+        // If a 401 Unauthorized error occurred, redirect to /signin.
+        // This will re-authenticate the user without showing a login page and a new token is issued.
+        logger.info("Attempting silent re-auth");
+        if (process && process.browser) {
+          Router.pushRoute(SIGN_IN_PATH);
+        } else {
+          // In server, if a 401 Unauthorized error occurred, redirect to /signin.
+          // This will re-authenticate without showing a login page and a new token is issued.
+          // Log out cookies so that the app will load unauthenticated if the re-auth doesn't work
+          if (options.req) options.req.logout();
+          if (options.res) {
+            options.res.writeHead(STATUS_FOUND, { Location: SIGN_IN_PATH });
+            options.res.end();
+          }
+        }
 
-      // eslint-disable-next-line no-console
-      console.error(`[Network error]: ${networkError}`);
+        return;
+      }
+      logger.error(`Unable to access the GraphQL API. Is it running and accessible at ${graphqlUrl} from the Storefront UI server?`);
     }
   });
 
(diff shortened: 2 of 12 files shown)


src/lib/apollo/initApollo.js
```diff
@@ -66,10 +83,9 @@ const create = (initialState, options) => {
     }
   }));
 
-  const httpLink = new HttpLink({ uri: `${graphqlUrl}`, credentials: "same-origin" });
+  const httpLink = new HttpLink({ uri: graphqlUrl, credentials: "same-origin" });
 
   return new ApolloClient({
-    connectToDevTools: process.browser,
```
aldeed: Apollo dev tools wasn't working for me anymore until I removed this. It automatically enables only for non-production, so it should be fine.

src/lib/apollo/withApolloClient.js
```diff
@@ -63,36 +57,12 @@ export default function withApolloClient(WrappedComponent) {
...
-                "Triggering token refresh via redirect flow");
-              res.writeHead(STATUS_FOUND, { Location: "/signin" });
-              res.end();
-              return {};
-            }
-            logger.error(`Unable to access the GraphQL API. Is it running and accessible at ${serverRuntimeConfig.graphqlUrl} from the Storefront UI server?`);
-          } else {
-            logger.error(`Error while running getDataFromTree: ${error}`);
-          }
-        }
```
aldeed: Essentially this same code was just moved to `initApollo.js`, where the browser redirect is happening, so that all of the logic is in one place (and because this catch was no longer being hit).

src/components/VariantItem/VariantItem.js
```diff
@@ -63,10 +63,16 @@ class VariantItem extends Component {
...
+    const className = classNames(variantButton, {
+      [activeVariant]: isActive
+    }, {
+      [soldOutVariant]: variantInventoryStatus && variantInventoryStatus.type === "SOLD_OUT"
+    });
+
     return (
       <ButtonBase
         disableRipple
-        className={classNames(variantButton, { [activeVariant]: isActive }, { [soldOutVariant]: variantInventoryStatus && variantInventoryStatus.type === "SOLD_OUT" })}
```
kieckhafer: Is this change meant to be in this PR? Seems out of place with the rest of the changes.

kieckhafer:
@aldeed I updated the snyk expiration date check here to fix that testing issue, at least until 2019-01-31. This looks pretty good, just had one question (above), but either way you answer that question it is still good to merge, just want to make sure you want that change here.

src/components/VariantItem/VariantItem.js
aldeed: It's not related, but it was an eslint warning so I fixed it

PR #472, opened 2019-01-04
refactor: navigation data comes from new NavigationTree

### LINKED ISSUE #446: Tags/NavBuilder: Navigation should come from new Nav object
As a Reaction storefront user, I can view the navigation built by the admin, so that I can see a nice custom Navigation.

As a followup to https://github.com/reactioncommerce/reaction/pull/4683, the navigation should come from the new navigation tree object.

## Dependency
https://github.com/reactioncommerce/reaction/pull/4683

kieckhafer:
Resolves #446
Impact: **minor**
Type: **refactor**

## Issue
As a Reaction storefront user, I can view the navigation built by the admin, so that I can see a nice custom Navigation.

## Solution
This is the UI portion of https://github.com/reactioncommerce/reaction/pull/4683
This PR updates the starterkit to use the `NavigationTree` to display navigation, rather than tags.

## Breaking changes
- Any store which uses tags as their navigation will need to build out a new Navigation Tree following the instructions in https://github.com/reactioncommerce/reaction/pull/4683
- Any custom code which was using the `primaryShopID` gql query will need to update this to use the `primaryShop` query, and use the `shop._id` from that data

## Testing
1. Follow the instructions in https://github.com/reactioncommerce/reaction/pull/4683 to create a NavigationTree
1. See that the UI displays this new tree as the navigation, instead of using existing tags
1. Change the language to `es` (or another language, if you've added translations) [here](https://github.com/reactioncommerce/reaction-next-starterkit/blob/refactor-kieckhafer-446-navigation/src/lib/data/getNavigationTree.js#L26), and see that the nav updates for items that have translations


src/containers/navigation/navigationTreeId.gql
```diff
@@ -0,0 +1,5 @@
+query navigationTreeIdQuery {
+  primaryShop {
+    defaultNavigationTreeId
+  }
```
aldeed: Currently `withShop` HOC does `primaryShopId` query followed by `shopQuery`, and now this adds a third query. But actually they can all be combined into one. We added `primaryShop` query in the API for this purpose, but never updated the client to use it.

So:
- Merge `shop.gql`, `primaryShopId.gql` and `navigationTreeId.gql` files into a single `primaryShop.gql` file, which should look like this one but with all of the fields from `shopQuery` added.
- Query `defaultNavigationTree` rather than `defaultNavigationTreeId` on `primaryShop`, and move all of the fields from `navigationQuery` into there. You then do not need `navigationQuery`.
- Update `withShop` to do just one query, `primaryShop`.
- Now you should no longer need `getNavigationTree` function. Instead, in `App`, you can get it from `shop.defaultNavigationTree` because it already uses `withShop`.

src/components/NavigationDesktop/NavigationItemDesktop.js
```diff
@@ -81,7 +81,7 @@ class NavigationItemDesktop extends Component {
       this.setState({ isSubNavOpen: !this.state.isSubNavOpen });
     } else {
       const path = this.linkPath();
-      Router.pushRoute(path, { slug: navItem.slug });
+      Router.pushRoute(path, { slug: navItem.navigationItem.data.url });
```
aldeed: I don't think having `slug` passed in here is needed or correct since the `path` from the server should now be the full URL? Just `Router.pushRoute(path)` should work.

src/components/NavigationDesktop/NavigationItemDesktop.js
```diff
@@ -90,10 +90,10 @@ class NavigationItemDesktop extends Component {
   };
 
   renderSubNav(navItemGroup) {
-    const menuItems = navItemGroup.subTags.map(({ node: navItem }, index) => (
+    const menuItems = navItemGroup.items.map((item, index) => (
       <MenuItem dense key={index}>
-        <Link onClick={this.onClose} route={`${this.linkPath(navItem)}`}>
-          <ListItemText primary={navItem.name} />
+        <Link onClick={this.onClose} route={`${this.linkPath(item)}`}>
+          <ListItemText primary={item.navigationItem.data.content[0].value} />
```
aldeed: I just noticed this `item.navigationItem.data.content[0].value`, which isn't what I had in mind when @dancastellon and I originally discussed the nav items development. I didn't catch this reviewing the GraphQL work, but I think we should fix it before merging this.

What it's doing is filtering the array by language. What I thought it would do is set `content` to the actual content for that language. So in the schema `content: [NavigationItemContent]` would be `content: String` instead.

Since we may need the full content array for operator UI, I'm thinking maybe we can leave what's there but add this to `NavigationItemData`:

```
contentForLanguage(language: String!): String
```

If we put the `language` param on that field rather than on the whole `defaultNavigationTree` query, I think it will be simpler and easier to understand what it's for.

Then here you would just do `item.navigationItem.data.contentForLanguage`

Thoughts @dancastellon ?

src/components/NavigationDesktop/NavigationItemDesktop.js
```diff
@@ -90,10 +90,10 @@ class NavigationItemDesktop extends Component {
   };
 
   renderSubNav(navItemGroup) {
-    const menuItems = navItemGroup.subTags.map(({ node: navItem }, index) => (
+    const menuItems = navItemGroup.items.map((item, index) => (
       <MenuItem dense key={index}>
-        <Link onClick={this.onClose} route={`${this.linkPath(navItem)}`}>
-          <ListItemText primary={navItem.name} />
+        <Link onClick={this.onClose} route={`${this.linkPath(item)}`}>
```
aldeed: The string template is unnecessary here. `route={this.linkPath(item)}`

src/lib/data/getAllTags.js
```diff
@@ -24,14 +24,15 @@ async function getTags(client, variables) {
 /**
  * @summary Gets all tags for the current shop from GraphQL and returns an array of them
  * @param {Object} client ApolloClient instance
+ * @param {Object} variables The query variables for this page
  * @returns {Object[]} Array of all tags
  */
-export default async function getAllTags(client) {
-  const { data: { primaryShopId } } = await client.query({ query: primaryShopIdQuery });
+export default async function getAllTags(client, variables) {
+  const { data: { primaryShop: shop } } = await client.query({ query: primaryShopQuery, variables });
```
aldeed: `variables` isn't being passed in so you can remove it

src/components/NavigationDesktop/NavigationItemDesktop.js
kieckhafer: I was a bit confused with this when setting it up. It makes sense to me to either have all the languages sent to the client, or only one, but not in an `array` like we currently have it.

aldeed:
@kieckhafer Yes I think to whatever is rendered for each one. Hopefully nobody is using classes for styling, but they might use them for click tracking or automated testing or something similar, so I think it makes sense to be on whatever element is clicked (is a link or has onClick).

src/components/NavigationDesktop/NavigationItemDesktop.js
dancastellon: @aldeed I just discussed this with @kieckhafer. I think it makes sense to remove the array filtering in the `content` resolver for the operator UI, and add that `contentForLanguage` resolver. I'll make the change.

src/containers/navigation/navigationItemFieldsFragment.gql
```diff
@@ -0,0 +1,11 @@
+fragment NavigationItemFields on NavigationItemData {
+  content {
+    language
+    value
+  }
```
aldeed: `content` is no longer needed, right? It could be a lot more data so we shouldn't request it if we don't need it.

src/components/Link/Link.js
```diff
@@ -68,13 +70,39 @@ class Link extends Component {
       ...props
     } = this.props;
 
+    let linkItemClassNames;
+
+    // If link is not a relative link, or if link should open in new window,
+    // then directly us an `a` tag, insted of the `NextLink` component
+    if (linkItem) {
+      const { navigationItem: { data } } = linkItem;
```
aldeed: I don't like coupling this component with the `linkItem` data structure. Can you just add `isUrlRelative` and `shouldOpenInNewWindow` props, and use the existing `className` prop? Then set those props in each place where this is used?

If that's too much duplicated logic, then you could make a `NavItemLink` component that wraps this one and have that one only accept `linkItem` prop.

src/components/NavigationDesktop/NavigationItemDesktop.js
```diff
@@ -116,20 +127,36 @@ class NavigationItemDesktop extends Component {
...
+                  </MenuList>
+                </Grid>
+              );
+            })}
           </Grid>
-          <Link className={classes.navigationShopAllLink} onClick={this.onClose} route={`${this.linkPath()}`} linkItem={navItem}>
+          <Link
+            className={classes.navigationShopAllLink}
+            onClick={this.onClose}
+            route={`${this.linkPath()}`}
```
aldeed: This doesn't need to be in a string template, and should `navItem` be passed in like the `href`?

aldeed:
This is ready to merge but we are waiting until the operator UI for editing the new nav is done. Otherwise storefront nav will be stuck at whatever is migrated with no way to update it.

machikoyasuda:
@kieckhafer now that https://github.com/reactioncommerce/reaction-next-starterkit/pull/496 has been merged, the Snyk errors should be resolved. if you rebase or merge in the latest changes, you should get passing tests! and then we can merge this.

machikoyasuda:
Now that https://github.com/reactioncommerce/reaction/pull/4936 is merged into `reaction/develop`, this PR is ready to merge.

PR #482, opened 2019-01-15
feat: add server req/res debug options

mikemurray:
Impact: **minor**
Type: **feature**

## Issue

Add debug options to help with identifying request/response headers in environments like staging.

## Solution

- Add env flags to enable debugging for req/res
- Log the req/res when available

## Breaking changes

none

## Testing

New env vars
```
DEBUG_REQUEST_HEADERS=false
DEBUG_RESPONSE_HEADERS=false
```

1. Copy above env vars to `.env` 
2. Set one or both to true
3. Load a page on the starter-kit
4. See messages logged starting with `Request header:` or `Response header:` followed by some output or `undefined`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,4 +1,6 @@
 CANONICAL_URL=http://localhost:4000
+DEBUG_REQUEST_HEADERS=false
+DEBUG_RESPONSE_HEADERS=false
 ENABLE_SPA_ROUTING=true
 EXTERNAL_GRAPHQL_URL=http://localhost:3000/graphql-alpha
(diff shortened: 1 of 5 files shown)


machikoyasuda:
@mikemurray Should we close this, since we now know what the issue was?

mikemurray:
@machikoyasuda We should close this PR.

spencern:
Did this get implemented? What was the issue that was solved?

mikemurray:
The purpose of this was to help @griggheo debug some things with an HTTP proxy.

machikoyasuda:
This PR made in response to Skipper and Staging core not properly working. But Grig figured it out without this.

PR #488, opened 2019-01-22
feat: Setup sitemap routes to get data from GQL server

impactmass:
Part of: reactioncommerce/reaction#4927
Impact: **major**
Type: **feature**

## Issue
Get Sitemap  data from GQL API and make it available on sitemap*.xml routes

## Testing
Same as reactioncommerce/reaction#4927, pasted below:
- Start reaction on this branch
- Generate Sitemaps by going to the "Shop" admin panel. Under "Options" tab, click "Refresh Sitemap"
- Start the starterkit app with this branch ([feat-4796-impactmass-sitemap-endpoint](https://github.com/reactioncommerce/reaction-next-starterkit/tree/feat-4796-impactmass-sitemap-endpoint)).
- Visit `{starterkitdomain}/sitemap.xml` e.g http://localhost:4000/sitemap.xml 
- You should see the sitemap xml on the page

### DIFF
--- a/src/sitemapRoutesHandler.js
+++ b/src/sitemapRoutesHandler.js
@@ -1,0 +1,51 @@
+const fetch = require("isomorphic-fetch");
+const config = require("./config");
+const logger = require("./lib/logger");
+
+/**
+ * @summary processes requests for sitemaps
+ * @param {Object} req - the request object
+ * @param {Object} res - the response object
+ * @param {Object} next - the next middleware function
+ * @returns {undefined}
+ */
+function sitemapRoutesHandler(req, res, next) {
+  if (req.originalUrl.startsWith("/sitemap") === false) {
+    return next();
+  }
+
+  res.setHeader("Content-Type", "text/xml");
+
+  const shopUrl = `${req.protocol}://${req.get("host")}`;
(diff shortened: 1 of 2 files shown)


src/sitemapRoutesHandler.js
```diff
@@ -0,0 +1,51 @@
...
+  const handle = req.originalUrl.replace("/", "");
+
+  const query = `
+    { sitemap(handle: "${handle}", shopUrl: "${shopUrl}") {
+      shopId
+      xml
+    } }
+  `;
+
+  return fetch(config.INTERNAL_GRAPHQL_URL, {
```
impactmass: Worth flagging here. Should this be `INTERNAL_GRAPHQL_URL` or `EXTERNAL_GRAPHQL_URL`? I made the decision based on the fact that the internal is the one added for nextjs `serverRuntimeConfig`; and I'm using it on the server.

kieckhafer: @aldeed ^^ see question above

aldeed: Should be correct. Internal should be used whenever it can be, external only when it's not on the Docker network, such as making a request from the browser.

PR #498, opened 2019-02-11
feat(tlp): add Featured as the default sort on tag listing page

### LINKED ISSUE #500: Tag Listing Page should be ordered by featured by default
https://github.com/reactioncommerce/reaction-next-starterkit/pull/498

Figure out a design for the Homepage so it does not conflict with the TLP default sort.

machikoyasuda:
closes #500 

- add `featured-desc` as an option in the tag listing page dropdown.
- add `featured-desc` as the default option in the tag listing page's `catalogItems` query, via the `UIStore.sortBy` variable.
- add snapshot test

## Dependency

Depends on https://github.com/reactioncommerce/reaction/pull/4980 being merged to `develop`

## Testing

- Check the Testing instructions from https://github.com/reactioncommerce/reaction/pull/4980
- Test that the Dropdown shows Features
- Test that clicking Featured in the Dropdown changes the sort
- Test that the default sort on first load is Featured

### DIFF
--- a/src/pages/tag.js
+++ b/src/pages/tag.js
@@ -160,7 +160,7 @@ export default class TagGridPage extends Component {
           pageSize={pageSize}
           setPageSize={this.setPageSize}
           setSortBy={this.setSortBy}
-          sortBy={sortBy}
+          sortBy={"featured-desc"}
         />
       </Fragment>
     );
--- a/src/pages/tag.js
+++ b/src/pages/tag.js
@@ -124,7 +124,7 @@ export default class TagGridPage extends Component {
       uiStore
     } = this.props;
     const pageSize = routingStore.query && routingStore.query.limit ? parseInt(routingStore.query.limit, 10) : uiStore.pageSize;
-    const sortBy = routingStore.query && routingStore.query.sortby ? routingStore.query.sortby : uiStore.sortBy;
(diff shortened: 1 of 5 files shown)


src/pages/tag.js
```diff
@@ -124,7 +124,7 @@ export default class TagGridPage extends Component {
       uiStore
     } = this.props;
     const pageSize = routingStore.query && routingStore.query.limit ? parseInt(routingStore.query.limit, 10) : uiStore.pageSize;
-    const sortBy = routingStore.query && routingStore.query.sortby ? routingStore.query.sortby : uiStore.sortBy;
+    const sortBy = routingStore.query && routingStore.query.sortby ? routingStore.query.sortby : "featured-desc";
```
machikoyasuda: @mikemurray What's the best way to set `featured-desc` to be the default sort only for the Tag Listing Page, and to use the URL param or the SortBy selector otherwise?

mikemurray: I believe `tag.js` is used for the tag pages so what you have above should work. What was `uiStore.sortBy` doing there, and was it needed for some reason?

machikoyasuda:
Closing this PR for now. The fact that the Homepage shares a Dropdown makes the design of this ticket slightly more complex than we have time for. Making an issue to complete this when a more detailed design is created.

PR #507, opened 2019-02-27
feat: add Orders to Account Profile

kieckhafer:
Expands upon #499 to add Orders data to profile page
Impact: **minor**
Type: **feature**

## Issue
Add Orders data to Account Profile

## Solution
- Split existing orders page into `profileAddressBook` and `profileOrders` pages. This is necessary due to the way we do pagination, our existing method broke when using `asPath` to display different components (`Orders` or `AddressBook`) on a single page based on the URL.
- Add `5` and `10` page size to our default pagination. Since we use the same component for CatalogItems / Tags, and now orders, and these return various dimensions of results we should allow more than just 20,60, or 100 for the results, as 20 is far too many `OrderCard`'s to scroll through
- Update `Thank You` page after an order is placed to use `OrderCard`, with the `Header` portion expanded by default (it's not expanded by default on the profile page).

## Breaking changes
None

## Testing

- Visit `/profile/orders` as a non-logged in user, and see a 404 / page not found
- Login as a user, create 6+ orders (to see pagination)
- Visit `/profile/orders` as a logged in user, see that orders show up and match our designs: https://zpl.io/V0ypYWR
- Move orders into different workflow statuses
- See orders are filtered as they should


kieckhafer:
@nnnnat Yes, that is expected if you haven't added the translations data. Instructions are here: https://github.com/reactioncommerce/reaction/pull/4981

Let me know if that fixes the issue. "new" is probably also wrong, it's probably not capitalized in your version. If you add them to the settings it'll change.

nnnnat:
@kieckhafer that makes sense I'll retest in a bit

aldeed:
@kieckhafer Regarding the copy issue Nat mentioned, it's still possible that people could eventually have long status labels when we allow customizing them. Can we do something with CSS, maybe `overflow-wrap: break-word;`, on the status badge to prevent it overlapping the date? cc @cassytaylor

package.json
```diff
@@ -48,6 +48,7 @@
     "compression": "^1.7.3",
     "cookie-parser": "^1.4.3",
     "cookie-session": "^2.0.0-beta.3",
+    "date-fns": "^1.30.1",
```
aldeed: Was there a decision to use `date-fns` instead of moment for new projects? It's fine with me but I'd like to see the decision recorded and communicated if it hasn't been.

src/components/OrderCard/OrderCard.js
```diff
@@ -0,0 +1,94 @@
...
+  };
+
+  renderFulfillmentGroups() {
+    const { order: { fulfillmentGroups } } = this.props;
+
+    return (
+      <Fragment>
+        {fulfillmentGroups.map((fulfillmentGroup, index) => (
+          <OrderCardFulfillmentGroup
+            key={`${index}`}
```
aldeed: It's always better to use a unique ID for `key` if there is one. In this case, you can use `fulfillmentGroup._id` as long as you request it in the GraphQL query.

src/components/OrderCardFulfillmentGroup/OrderCardFulfillmentGroup.js
```diff
@@ -0,0 +1,138 @@
...
+          />
+        </Grid>
+      );
+    }
+
+    return null;
+  }
+
+  onTrackShipmentButtonClick() {
+    // TODO: What do we do to track a shipment? Link to a specific provider website?
```
aldeed: Maybe I'll add `trackingUrl` prop on `fulfillmentGroup` in my tracking PR, and then you can render that in a link here? Seems better than recording the carrier and trying to dynamically build links to different carriers.

package.json
kieckhafer: We had a chat in Slack about a light(er)weight solution for dates, considering we need very little. Made a DR and tagged you in Notion.

src/components/OrderCardFulfillmentGroup/OrderCardFulfillmentGroup.js
kieckhafer: Yeah, that sounds like a good idea.

kieckhafer:
@aldeed did not update the long name on the Badge yet. create an issue here to have design take a look since it shouldn't' be a blocker: https://github.com/reactioncommerce/reaction-next-starterkit/issues/515

PR #514, opened 2019-03-06
remove blocking code to allow starterkit to start

### LINKED ISSUE #513: Unable to create starterkit docker container, therefore unable to run starterkit
Type: **breaking**

**Describe the bug**
Unable to start `starterkit`, with the following error thrown:

```
ERROR: unsatisfiable constraints:
  curl-7.61.1-r2:
    breaks: world[curl=7.61.1-r1]
ERROR: Service 'web' failed to build: The command '/bin/sh -c apk --no-cache add bash=4.4.19-r1 curl=7.61.1-r1 less=530-r0 vim=8.1.0115-r0' returned a non-zero code: 1
```

**To Reproduce**
1. Clone `reaction-platform`
1. run `make`
1. See that all containers starts, except starterkit, which throws the above error.

**Expected behavior**
Starterkit starts when `make` is run

**Screenshots**
1__reaction-platform__bash_

 - OS: macOS 10.14.3
 - Browser: n/a

**Possible solution**
Removing lines 59 & 60 in the `Dockerfile` solve the issue: https://github.com/reactioncommerce/reaction-next-starterkit/blob/develop/Dockerfile#L59

kieckhafer:
Resolves #513 
Impact: **breaking**
Type: **bugfix**

## Issue
Starterkit docker container was throwing errors, and not being created:
```
ERROR: unsatisfiable constraints:
  curl-7.61.1-r2:
    breaks: world[curl=7.61.1-r1]
ERROR: Service 'web' failed to build: The command '/bin/sh -c apk --no-cache add bash=4.4.19-r1 curl=7.61.1-r1 less=530-r0 vim=8.1.0115-r0' returned a non-zero code: 1
```

## Solution
Remove two lines of code contributing to this error. It seems like this code might have been transferred over from https://github.com/reactioncommerce/reaction-cli, and isn't working in its current form.

## Testing
1. Clone `reaction-platform`
1. run `make`
1. See that `starterkit` starts with no errors

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,11 +54,6 @@ LABEL maintainer="Reaction Commerce <engineering@reactioncommerce.com>" \
       com.reactioncommerce.docker.git.sha1=$GIT_SHA1 \
       com.reactioncommerce.docker.license=$LICENSE
 
-# Get versions to pin with this command:
-# apk list bash curl less vim | cut -d " " -f 1 | sed 's/-/=/' | xargs
-RUN apk --no-cache add bash=4.4.19-r1 curl=7.61.1-r1 less=530-r0 vim=8.1.0115-r0
-SHELL ["/bin/bash", "-o", "pipefail", "-o", "errexit", "-u", "-c"]
-
 # Because Docker Compose uses a volume for node_modules and volumes are owned
 # by root by default, we have to initially create node_modules here with correct owner.
 # Without this Yarn cannot write packages into node_modules later, when running in a container.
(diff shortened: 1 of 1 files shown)


focusaurus:
OK here's the full context here. There's a Dockerfile best practice linter called hadolint and it has the following rule/recommendation:

https://github.com/hadolint/hadolint/wiki/DL3018

This says you should pin your apk package version numbers so your image builds are more consistent. However, I think I probably misunderstood what is workable. I am pinning hyper-specific versions and it seems those versions disappear at some point as they are superseded with security updates, bug fixes, etc.

So I think we want either:

- Don't pin versions at all, disable the DL3018 lint
  - Totally fine to merge this way to unblock folks while we discuss
- Pin less-specific versions like `curl=7.61.*`
  - Maybe there's some benefit here? 

Any docker nerds have opinions? @rosshadden @ticean ?

PR #523, opened 2019-04-09
chore: ignore snyk js-yaml vuln for 30 days

focusaurus:
Impact: **minor**
Type: **chore|security**

## Issue

We have a snyk vuln deep in our dep tree that is non-trivial to avoid with npm dependency updates at this time.

## Solution

Delay 30 days and hope some new deps show up that don't break when we update.

## Breaking changes
N/A

## Testing
N/A

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,4 +1,13 @@
 # Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
-version: v1.13.3
+version: v1.13.4
 patch: {}
-ignore: {}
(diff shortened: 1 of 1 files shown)


focusaurus:
Y U NO PASS?!?

focusaurus:
Oh dang there are 2 different js-yaml vulns. Yay YAML is so great!

focusaurus:
OK this is good to go now I think @nnnnat . The license compliance check is maybe choking on the fact that I amended my commit and force pushed, but it's not relevant to this change anyway.

PR #531, opened 2019-05-31
chore: update yarn.lock to resolve snyk js-yaml vuln

focusaurus:
- Change apollo-related package.json semver ranges to tilda. Otherwise we get stuff too new and that causes unit test failures with `Invariant Violation: Ambiguous GraphQL document: contains 2 operations`
- update yarn.lock to fix js-yaml vulnerability
- remove ignored vulnerabilities as they should now be resolved


focusaurus:
OK yay this finally passed circleci.

focusaurus:
@spencern If we leave the carets for apollo stuff, an update breaks the unit tests with `Invariant Violation: Ambiguous GraphQL document: contains 2 operations`. The switch to tilde is a smaller range and prevents us jumping minor versions at least until we have time to actually deal with the new graphql query invariant check. I tried to change as little as possible. If I were to re-think our semver approach, I'd suggest we always use `7.6.x` syntax because 

1. We want to allow patch fixes mostly for security updates
2. We want to avoid minor updates because they often are breaking even though they are not supposed to be 
3. The `.x` syntax for me personally is way clearer. I have to always double check the docs to understand the tilde and caret rules. However, in rare cases the `.x` might need to be replaced when we require `>=` a specific patch version.

spencern:
@focusaurus 
I agree with your decision to change as little as possible in this particular PR. I'd be interested in seeing a DR/Proposal for rethinking our semver approach. My experience matches yours that minor versions too frequently do contain breaking changes.
I don't have strong attachment to the existing way of specifying versions (`~ 7.6.1`) but I'd want to hear others opinions on that before making that change.

PR #532, opened 2019-05-31
fix: prettier config was in the wrong place

focusaurus:
Fixes this warning: 

```
● Validation Warning:

  Unknown option "prettier" with value {"arrowParens": "always"} was found.
  This is probably a typing mistake. Fixing it will remove this message.

  Configuration Documentation:
  https://jestjs.io/docs/configuration.html
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,9 +129,6 @@
       "!**/vendor/**"
     ],
     "coverageDirectory": "reports/coverage",
-    "prettier": {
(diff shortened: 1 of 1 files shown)


kieckhafer:
`snyk` is failing here, but it was updating in another PR a few minutes ago, so merging this assuming it is fixed based on that PR.

PR #533, opened 2019-05-31
WIP: Run docker container using particular uid:gid

focusaurus:
Impact: **breaking**
Type: **feature|bugfix**

## Issue

On linux, the docker container runs as uid 1000, which in many cases is not the same as the developer's host OS uid. This can cause mismatches in ownership of files that are shared between the host and the container via docker mounts. Ultimately this surfaces as filesystem errors (`EACCESS`, `permission denied`, etc) at various times including when trying to run `yarn`. Mac has automatic mapping that avoids this issue largely.

## Related issues

- https://github.com/reactioncommerce/reaction-next-starterkit/pull/530

## Solution

Configure docker-compose to run the container as the same UID as the host OS user. This is initialized in the `bin/setup` script.

## Breaking changes

Existing files both in the host filesystem and in the docker volumes may be owned by uid 1000 (for this repo). That will probably cause some errors.

For docker volumes: `docker-compose down -v` should resolve them. (at the cost of blowing away your node_modules)

For the host OS: `sudo chown -R "${USER}" ~/.cache/yarn-offline-mirror` should fix it.

## Testing

- Without having run `bin/setup` start the storefront. It should default to empty string and still work barring ownership issues
- Run `bin/setup` and then restart the storefront. It should work.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -4,27 +4,47 @@ __dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 env_file=${__dir}/../.env
 env_example_file=${__dir}/../.env.example
 
-function main {
+function main() {
   set -e
 
   add_new_env_vars
+  add_reaction_user
 }
 
-function add_new_env_vars {
+function add_new_env_vars() {
   # create .env and set perms if it does not exist
-  [ ! -f "${env_file}" ] && { touch "${env_file}" ; chmod 0600 "${env_file}" ; }
+  if [[ ! -f "${env_file}" ]]; then
+    touch "${env_file}"
+    chmod 0600 "${env_file}"
+  fi
 
-  export IFS=$'\n'
-  for var in $(cat "${env_example_file}"); do
-    key="${var%%=*}"     # get var key
+  while IFS=$'\n' read -r var; do
+    key="${var%%=*}"        # get var key
     var=$(eval echo "$var") # generate dynamic values
 
     # If .env doesn't contain this env key, add it
-    if ! $(grep -qLE "^$key=" "${env_file}"); then
+    if ! grep -qLE "^$key=" "${env_file}"; then
       echo "Adding $key to .env"
-      echo "$var" >> "${env_file}"
+      echo "$var" >>"${env_file}"
     fi
-  done
+  done <"${env_example_file}"
+}
+
+function add_reaction_user() {
+  if grep -qLE "^REACTION_USER" "${env_file}"; then
+    # line is already present, leave unchanged
+    return
+  fi
+  # line is not present
+  if [[ -n "${REACTION_USER}" ]]; then
+    # if set in environment, just add the name but no = and no value
+    line="REACTION_USER"
+  else
+    # if unset in environment, default to the current user's uid
+    line="REACTION_USER=$(id -u):$(id -g)"
+  fi
+  echo "Adding ${line} to .env"
+  echo "${line}" >>"${env_file}"
 }
 
 main
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,6 +26,7 @@ services:
       auth.reaction.localhost:
     ports:
       - 4000:4000
+    user: ${REACTION_USER:-1000:1000}
     volumes:
       - $HOME/.cache/yarn-offline-mirror:/home/node/.cache/yarn-offline-mirror
       - web-yarn:/home/node/.cache/yarn


focusaurus:
> Some namespacing for the var name might be good, like REACTION_HOST_UID?

:+1: 
----

>I propose that we don't tightly couple the container's user with the one that's currently running bin/setup.
> the id for the remote Docker for Mac should be 1000

I'll bust out my mac and do some testing, but my understanding at the moment was it didn't really matter what userid is used in the container on mac. The translation maps it back to the host uid that launched the container so as I understand it essentially any value should work the same. I'll test out to confirm though (tomorrow).

I did some testing around default values.

**Yes**, `docker-compose run` will interpolate variable syntax in the `docker-compose.yml` file and it supports defaults

```sh
╭ /tmp/20190605-003508 plyons@leop
╰○ echo  $REACTION_HOST_UID

╭ /tmp/20190605-003508 plyons@leop
╰○ grep user docker-compose.yml
    user: ${REACTION_HOST_UID:-1111:1111}
╭ /tmp/20190605-003508 plyons@leop
╰○ docker-compose run myservice id
uid=1111 gid=1111 groups=1111
/tmp/20190605-003508 plyons@leop
╰○ REACTION_HOST_UID=2222:2222 docker-compose run myservice id
uid=2222 gid=2222 groups=2222
```

**No**, `docker-compose run` does **not** interpolate variable syntax in the `.env` file though. But it will accept a plain value without an equals sign and grab the value from the environment if set.

```sh
╭ /tmp/20190605-003508 plyons@leop
╰○ cat .env 
FOO=${SOMEVAR:-defaultfoo}
BAR
╭ /tmp/20190605-003508 plyons@leop
╰○ SOMEVAR=explicitvalue BAR=somebar docker-compose run myservice env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=2fd2f6b7c571
TERM=xterm
FOO=${SOMEVAR:-defaultfoo}
BAR=somebar
HOME=/root
```

> A separate script for setting HOST_UID or a flag in bin/setup could help with that

As to an explicit script to set this, or a flag to `bin/setup` I don't feel super strongly but here's my thinking:

- `bin/setup` should compute this and put it into .env if not already there
  - It will be a good value nearly always
  - I think it's fine to do this on both linux and mac
- As needed, users can edit the file and change the value.
  - Once present, the value will not be overridden (standard setup behavior)
  - A script to change this simple text file seems like more support/doc overhead then just telling a user to edit the file
  - If for some reason you really want to change your UID or do it as an env var, set it to just `REACTION_HOST_UID` (no value) in `.env` and export it in your shell and you're good for all projects

focusaurus:
OK I tested on mac and it looks like essentially the uid in the running container doesn't really matter. Anything will work but the filesystem owner will get changed back to the host uid automatically.

```
$ REACTION_HOST_UID=2222:2222 docker-compose run mytestservice touch /opt/reaction/src/fstest1
$ ls -ln
-rw-r--r--  1 501  0    0 Jun  5 09:35 fstest1
$ echo file-content > not-readable
$ chmod 600 not-readable 
$ REACTION_HOST_UID=2222:2222 docker-compose run mytestservice cat /opt/reaction/src/not-readable
file-content
```

aldeed:
@focusaurus In the Dockerfiles, we use `--chown=node` and `USER node` (https://github.com/reactioncommerce/reaction-next-starterkit/blob/develop/Dockerfile#L111), which I thought was the solution to this issue (at least on Mac). How do your changes relate? Are all of them necessary? Do your changes map OS user to `node` user in the container, but we still need to ensure `node` user is used rather than root?

focusaurus:
@aldeed That is not the solution to the issue on linux, it's part of the problem. I think inside the container the issue almost never surfaces because most files/directories and world readable in the container (which is fine and good) and we usually don't write directly to files in the container, only to volumes. On mac that seems to mostly work fine but on linux it becomes a problem when the container uid doesn't match and the mounted directories aren't world writeable, which they typically would not be. So this change makes it so the userids match and thus if the user can write files to a directory normally, we'll be able to write to that directory from within a container via a volume mount and all will work OK. The `--chown=node` and `USER node` lines are still good and needed, and yes running as non-root is still what we want.

aldeed:
@focusaurus One note: the last time we made breaking changes like this it led to several weeks of confusion and wasted time, even though the change was communicated internally. People missed the notice, or didn't understand it, or didn't think it applied to them, or whatever. This is not to say it shouldn't be done, but I would lean toward over-communicating this change with very specific instructions for rebuilding locally.

janus-reith:
I'm wondering if this fix just applies to the starterkit?
I believe there were similar permisssion issues on the meteor container, although they were less common than on the starterkit.

aldeed:
Yes, I assume we'll want to do this for all repos as well as the project generator template

focusaurus:
@aldeed I agree about overcommunicating. I'd also like to potentially script some more specific fixes and either ship them with this so tech support 90% case is "run scripts/fix-ownership.sh" or something like that. Let's get @manueldelreal involved with testing early and maybe reproducing some of the various situations we will encounter out in the developer base.

focusaurus:
Discussed some more with @ticean. Renamed the variable to `REACTION_USER`. It's value can be a numeric userid or `uid:gid`. Made a change such that the logic in `bin/setup` is

- If `.env` already has`REACTION_USER`, do nothing
- If `.env` does not have it, but it's in the environment, put `REACTION_USER` with no value in the `.env` file, and `docker-compose` will use the value from the environment
- if `.env` does not have it, and it's not in the environment, put a default value in `.env` which is the current user's uid:gid

focusaurus:
OK @manueldelreal did some testing and at the moment I don't understand where the ownership of docker data volume mounts is controlled. I need to do some more research and figure out if it's possible to control that.

focusaurus:
Some interesting attempts in [schmidigital/permission-fix](https://github.com/schmidigital/permission-fix), but as of this moment I think this PR as currently proposed is not viable.

focusaurus:
More research: https://denibertovic.com/posts/handling-permissions-with-docker-volumes/

focusaurus:
OK my testing has revealed there's more going on here at runtime and latest thinking is I'll need a runtime entrypoint script, which will be a different approach from this PR and require time to develop and test, so I'm going to close this and open a new one referencing this when I have something ready.

PR #550, opened 2019-07-02
Add 302 redirects to env.EXTERNAL_GRAPHQL_URL

### LINKED ISSUE #549: Redirect graphql playground (graphiql) to core
# Feature Name

### Summary description

I often can't remember where we host our graphql playground (graphiql). Since it's a web app type thing, I often guess it is served by the storefront. In fact it's served by the graphql API server. I'd like to suggest a series of developer-friendliness 30X redirect routes in storefront to make this "just work" even if I forget:

- `/graphql-beta` (actual path used by core)
- `/graphiql` (actual name of the 3rd party project)
- `/graphql` (intuitive guess)
- `/graphql-alpha` (path used until v2.0.0-rc.12)

These should all redirect to: `http://reaction.api.localhost:3000/graphql-beta`, which comes from the `EXTERNAL_GRAPHQL_URL` env var.

Improves developer experience by eliminating potentially frustrating dead ends and error pages.

cshepherd:
Signed-off-by: Christopher Shepherd <christopher@reactioncommerce.com>

Resolves #549 
Impact: **minor**
Type: **feature**

## Issue
Adds 302 redirects for a few paths via express.js

## Solution
Added redirects at the express.js level rather than the next.js router because I don't see a downside to it, and it's easy to express in just a few lines of code.

## Breaking changes
none

## Testing
1. curl -i http://localhost:4000/graphiql (or /graphql, or /graphql-alpha or /graphql-beta)
2. Observe HTTP 302 redirect to EXTERNAL_GRAPHQL_URL

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -42,7 +42,7 @@ app
 
     // add graphiql redirects to EXTERNAL_GRAPHQL_URL
     server.get(["/graphiql", "/graphql-beta", "/graphql-alpha", "/graphql"], (req, res) => {
-      res.redirect(config.EXTERNAL_GRAPHQL_URL);
+      res.redirect(301, config.EXTERNAL_GRAPHQL_URL);
     });
 
     // apply to routes starting with "/sitemap" and ending with ".xml"
--- a/src/server.js
+++ b/src/server.js
@@ -41,8 +41,8 @@ app
     configureAuthForServer(server);
 
     // add graphiql redirects to EXTERNAL_GRAPHQL_URL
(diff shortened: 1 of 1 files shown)


cshepherd:
> LGTM! My only thought is maybe `301 MOVED PERMANENTLY` is the more pedantic/correct code?

The redirect is permanent and 301 is therefore correct, but browsers will cache a 301 as long as possible in the absence of a Cache-Control: header, and someday 'reaction/graphql-beta' will probably become 'reaction/graphql'. Perhaps 301 with "Cache-Control: max-age=86400" header? Or am I overthinking this?

PR #573, opened 2019-09-04
feat: match userid between host and container

focusaurus:
Impact: **major**
Type: **feature|bugfix**

## Issue

On linux, the docker container runs as user `node` (uid 1000), which in many cases is not the same as the developer's host OS uid. This can cause mismatches in ownership of files that are shared between the host and the container via docker volume mounts. Ultimately this surfaces as filesystem errors (`EACCESS`, `permission denied`, etc) at various times including when trying to run `yarn`, do builds, etc.

**Why doesn't bug  affect all linux users?**

On many linux distributions, the first non-system user is assigned uid 1000. That's why in our docker container the `node` user has uid 1000, it's just a default. That's also why many developer's have uid 1000 as well: it's the default for the first non-system user on their distribution and their user is the first one they add when doing their OS install. Because of that default and coincidence many users avoid this issue because their uid just happens to match the one we use. (This was the case for me personally and why it is harder for me to encounter/reproduce these issues)

**Why doesn't this affect macOS users?**

Docker for Mac has automatic mapping that avoids this issue largely.

**How does this apply to reaction core? Other repos?**

If testing proves successful in storefront, I believe this pattern would apply to reaction core and essentially every docker-based service we have that makes use of volume mounts, which is probably most but not all of them. I started with core initially as a fix there would have the widest impact, but the docker setup there is significantly more complex and with the slow startup time I switched to storefront as an easier first project to tackle.

## Related issues

- Storefront PR attempt from May 2019 (closed unmerged when problem was not solvable without a root entrypoint script)
  - https://github.com/reactioncommerce/example-storefront/pull/533#pullrequestreview-
- many users in the community hitting a version of this issue starting in Oct 2018
  - Specifically: `Error creating Reaction config file: /opt/reaction/src/.reaction/config.json`
- Me hitting a loosely-related issue in storefront
- community user issue
  - https://github.com/reactioncommerce/reaction/issues/5510

## Solution

The key elements of the solution are:

- Create an env var mechanism so inside docker we can know the host uid
- Start the docker process as root
- During container startup, modify the `node` user account in the container to have the matching uid
change ownership and permissions on all necessary directories to the matching uid
- `su-exec` to `node` in the container then proceed to launch the application
- Provide a `./bin/fix-volumes` script that can be run at any time that will `chown`/`chmod` all volume mount directories properly and should be a 1-stop fix to this entire category of errors

There are a ton of unix heavy details in here we should scrutinize during code review.

Some things to note about the solution (pending QA testing)

- The goal here is to have things "just work" in all cases
- File owners and permissions on the host filesystem will be changed when running `fix-volumes`
  - There's potential here for surprise or confusion in the user base, and perhaps "hey don't do that!"
- I believe the solution will handle pre-existing files in the volume mount directories with assorted owner/permission combinations and it should force them all to be correct, but I think there's a lot of testing surface here

## Breaking changes

I don't think anything here would count as "breaking" but the changed ownership/permission of host files could be surprising/unexpected as noted above.

## Testing

- Try a fresh `git clone` of this example-storefront branch, `./bin/setup`, and `docker-compose up`
  - Try on mac, linux with userid !=1000, linux with userid=1000
  - Directories of interest include
    - `$HOME/.cache/yarn-offline-mirror`
    - `$HOME/.cache/yarn`
    - `example-storefront/node_modules`
    - `example-storefront/build`
- Note you may want to create a new user in your linux host for this to force non-1000 uid. On linux mint I was able to do this via the users & groups GUI (or you can use `adduser` CLI) and once I did `sudo adduser plyons2 docker` the new user could use  docker. Then I `sudo su - plyons2` to get a shell with that user for testing.
- Verify when the application finally loads that it is not running as root
  - `docker-compose run web ps -ef`
  - You will see some early `root` processes then a switch to `node` for the `bin/start`

**Example good output**
- Your container id may vary
- If everything is running as `root`, that's a bug

```
docker exec --interactive --tty rc-storefront_web_1 ps -ef
PID   USER     TIME  COMMAND
    1 root      0:00 su --preserve-environment --command /usr/local/src/reactio
   23 node      0:00 sh /usr/local/src/reaction-app/bin/start
   67 node      0:00 node /opt/yarn-v1.13.0/bin/yarn.js dev
   88 node      0:15 /usr/local/bin/node ./src/server.js
   99 root      0:00 ps -ef
```

### DIFF
--- a/.reaction/entrypoint.sh
+++ b/.reaction/entrypoint.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+# Please Use Google Shell Style: https://google.github.io/styleguide/shell.xml
+
+# ---- Start unofficial bash strict mode boilerplate
+# http://redsymbol.net/articles/unofficial-bash-strict-mode/
+set -o errexit  # always exit on error
+set -o errtrace # trap errors in functions as well
+set -o pipefail # don't ignore exit codes when piping output
+set -o posix    # more strict failures in subshells
+# set -x          # enable debugging
+
+IFS=$'\n\t'
+# ---- End unofficial bash strict mode boilerplate
+
+cd "$(dirname "${BASH_SOURCE[0]}")/.."
+uid=$(echo "${REACTION_USER:-1000:1000}" | cut -d : -f 1)
+usermod --uid "${uid}" --non-unique node |& grep -v "no changes" || true
+command=("./bin/start")
+if [[ $# -gt 0 ]]; then
+  command=($@)
+fi
+unset IFS
+exec su-exec node ${command[*]}
--- a/.reaction/fix-volumes.sh
+++ b/.reaction/fix-volumes.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+
+# Please Use Google Shell Style: https://google.github.io/styleguide/shell.xml
+
+# ---- Start unofficial bash strict mode boilerplate
+# http://redsymbol.net/articles/unofficial-bash-strict-mode/
+set -o errexit  # always exit on error
+set -o errtrace # trap errors in functions as well
+set -o pipefail # don't ignore exit codes when piping output
+set -o posix    # more strict failures in subshells
+# set -x          # enable debugging
+
+IFS=$'\n\t'
+# ---- End unofficial bash strict mode boilerplate
+
+cd "$(dirname "${BASH_SOURCE[0]}")/.."
+uid=$(echo "${REACTION_USER:-1000:1000}" | cut -d : -f 1)
+volumes=(
+  ../node_modules
+  ./build
+  /home/node/.cache/yarn-offline-mirror
+  /home/node/.cache/yarn
+)
+for dir in ${volumes[*]}; do
+  printf "Fixing volume ${dir} (uid ${uid})…"
+  mkdir -p "${dir}"
+  chown -R node "${dir}"
+  chmod -R a+r,u+rw "${dir}"
+  echo "✓"
+done
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,7 +54,9 @@ LABEL maintainer="Reaction Commerce <engineering@reactioncommerce.com>" \
       com.reactioncommerce.docker.git.sha1=$GIT_SHA1 \
       com.reactioncommerce.docker.license=$LICENSE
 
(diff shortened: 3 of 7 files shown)


focusaurus:
There's an important performance trade-off to consider here. I'm currently doing some recursive filesystem operations at startup, which take some time. On linux, it's typically pretty short (like 1s maybe) but in cases where people have way more files lying around than we typically see, it could get slower. On macos, it may be a lot slower due to docker filesystem performance. It might be worth trying to skip these entirely if the host OS is macOS (would need to research if we can detect that and how). Another possibility would be to split out the "just make it work" changes to a separate script we don't run by default during startup but make it easy to run ad hoc when you hit errors.

focusaurus:
FYI I tested commit 743edfc  on mac and the same code designed for linux seems to work OK and be reasonable on mac too. We may want to try to skip it on mac for performance reasons but we'll see.

focusaurus:
OK I converted it to an ad-hoc script and swapped in `su-exec` which is much nicer than normal `su`. So now there are 2 interesting changes here:

1. The container starts as root, changes the `node` user's uid to match the host uid, then starts the app as `node` (via `./bin/start-root`)
2. As needed, users can run `./bin/fix-volumes` from the host and it should fix any permission errors

focusaurus:
OK I think this has reached a 2nd milestone. Going to rebase and clean up my commits and convert to real PR.

focusaurus:
For reference a [blog post](https://denibertovic.com/posts/handling-permissions-with-docker-volumes/) with basically our technique of an entrypoint script (but we change the `node` user's uid instead of creating a new user).

rosshadden:
Should we create a new user instead, like they do?

focusaurus:
OK I learned some new tricks and I think I have a new approach including:

- use `owner=$(stat -c "%u:%g" .)` to discover the owner of the example-storefront directory from within the container
  - Have to check macos to see what that does and whether we need to handle separately
  - This should eliminate the REACTION_USER stuff altogether
- One call to `stat` each volume mount, if they are already OK, proceed with app start. If not, do the `chown -R` to fix it
  - This should be a very acceptable robustness/speed tradeoff to have us just do this at startup always

focusaurus:
Closing in favor of https://github.com/reactioncommerce/example-storefront/pull/576

PR #599, opened 2019-10-10
refactor: remove deprecated api field

kieckhafer:
Related to https://github.com/reactioncommerce/reaction/issues/5703
Impact: **minor**
Type: **refactor**

## Issue
`currentQuantity` will be deprecated in Reaction 3.0. It has already been replaced with `inventoryAvailableToSell` in the `v2.x` branches, this is just an official removal of the field.

## Solution
Remove all references to `currentQuantity` in the storefront.

## Breaking changes
Any custom code using `currentQuantity` will need to be updated prior to using Reaction `v3.0.0`


aldeed:
@kieckhafer There are some failing tests (snapshots), and the reason is because some components from component lib are still expecting `currentQuantity` prop. You will have to request `inventoryAvailableToSell` from GraphQL and then you can either map it to `currentQuantity` before passing into those components, or update them in comp lib package to accept either `inventoryAvailableToSell` or `currentQuantity` props.

aldeed:
Note: After we merge this, merge into `release-v3.0.0` branch as well.

PR #623, opened 2019-12-09
chore: use published image for docker-compose

aldeed:
Impact: **minor**
Type: **chore**

## Changes
When you clone this branch and run `docker-compose up`, it will now pull and use the latest published `reactioncommerce/example-storefront:release-v3.0.0` image from DockerHub instead of `node-dev`. Host files are not linked in, NPM deps are not reinstalled, and NextJS will run in the faster production mode unless you set `NODE_ENV=development` in your `.env` file.

## Breaking changes
Breaking only for devs who will now need to run `ln -s docker-compose.dev.yml docker-compose.override.yml` to get the previous behavior.

## Testing
- Verify `dc up` starts quickly and in prod mode. (You should not see anything about compiling pages in the logs.)
- After `dc down`, run `ln -s docker-compose.dev.yml docker-compose.override.yml` and then `dc up` again. Verify it now starts in dev mode and reloads when you change files.
- After `dc down`, run `rm docker-compose.override.yml` to remove the override file. It should be back to running in prod mode when you do `dc up`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,4 +32,4 @@ src/.next/*
 build
 /reports
 
-package-lock.json
+docker-compose.override.yml
--- a/bin/start-prod.sh
+++ b/bin/start-prod.sh
@@ -7,6 +7,7 @@ set -o pipefail # don't ignore exit codes when piping output
 IFS="$(printf "\n\t")"
 
 cd "$(dirname "$0")/.."
-yarn install
-./bin/create-hydra-client.sh
-yarn start:dev
+printf "Creating hydra client…"
+./bin/wait-for.sh "${OAUTH2_HOST}:${OAUTH2_ADMIN_PORT}"
+./bin/create-hydra-client.js
+node .
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -1,0 +1,26 @@
+# This docker-compose file is used to run the project in Docker for development.
+# The local files are mounted into the created container.
+#
+# Usage:
+#  ln -s docker-compose.dev.yml docker-compose.override.yml
+#  docker-compose up [-d]
+#
+# To go back to running the published image:
+#  rm docker-compose.override.yml
+
+version: '3.4'
+
+services:
+  web:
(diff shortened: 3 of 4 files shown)


focusaurus:
I'll give this a local test this afternoon.

focusaurus:
I seem some syntax errors but those aren't related to the docker mechanics here.

```
web_1  |  error  in ./src/pages/_error.js
web_1  |
web_1  | Syntax Error: SyntaxError: /usr/local/src/app/src/pages/_error.js: Support for the experimental syntax 'decorators-legacy' isn't currently enabled (23:1):
web_1  |
web_1  |   21 | });
web_1  |   22 |
web_1  | > 23 | @withStyles(styles)
web_1  |      | ^
web_1  |   24 | export default class Error extends Component {
web_1  |   25 |   static propTypes = {
web_1  |   26 |     classes: PropTypes.object,
web_1  |
web_1  |
web_1  |  @ multi ./pages/_error.js
```

focusaurus:
Wait hmm. It looks like the server starts and THEN does try to do some compiling:

```
web_1  | Server started ! ✓
web_1  |
web_1  |       http://localhost:4000
web_1  |       Press CTRL-C to stop
web_1  |
web_1  |  WAIT  Compiling...12:39:29 AM
web_1  |
web_1  | [12:39:29 AM] Compiling client
web_1  | [12:39:29 AM] Compiled client in 51ms
web_1  |  ERROR  Failed to compile with 2 errors12:39:29 AM
web_1  |
web_1  |  error  in ./src/pages/_app.js
w
```
 
Is that right for prod mode?

aldeed:
@focusaurus It shouldn't be doing that when you don't have the override file. Can you verify you don't have `NODE_ENV=development` in `.env`? That would also cause it. Even the prebuilt NextJS server will run in dev mode when `NODE_ENV` is `development`, which is always confusing to me.

focusaurus:
Yes, I had `NODE_ENV=development` in my `.env`. After commenting that out it behaves properly. Should we add a warning to the prod image to catch this? What percentage of our user base would hit the same scenario I did?

PR #637, opened 2019-12-30
Change password and logout

### LINKED ISSUE #574: Remove id param from logout endpoint
### Rationale for why this is necessary
The logout endpoint takes the userId in the url param. Meaning that you can pass any known ID to log out a user. Even though we have the id in an encoded form, we should remove the id param.

### Work:
Use `/logout` without the id param. Find alternative to retrieve the current user and the issue to logout api call.

aldeed:
Resolves #574 
Impact: **minor**
Type: **feature**

## Breaking changes
These changes are tied to changes in Reaction Identity. As long as both services are updated as well as any environment variables, nothing should break.

This requires additional scope and options for the Hydra client, but I've included server startup code that will auto-update the Hydra client as necessary.

## Testing
Test with the following PR branches:
- https://github.com/reactioncommerce/reaction/pull/5991
- https://github.com/reactioncommerce/reaction-hydra/pull/42
- https://github.com/reactioncommerce/reaction-identity/pull/12

### Reset Password
Prerequisite: Configure emailing on the API so that you'll get the password reset email.

1. Create an account
2. Sign out
3. Click Sign In, but then click Forgot Password
4. Enter the email address of the account you created and click the button.
5. In the email, click the link.
6. Set a new password
7. Log in with the new password to confirm it worked.

### Change Password
1. Create an account or use already created.
2. Sign in
3. From account menu, choose Change Password.
4. You should be taken to the Change Password form on Reaction Identity with the email address pre-filled.
5. Enter your current and new password.
6. After the change, you should be automatically redirected back to the same storefront page you were on when you clicked Change Password.
7. Sign out and back in with the new password to verify it worked.

### Logout
If you've tested the other changes, you've been testing logout. The flow is different behind the scenes (standard OpenID Connect Logout Flow, more secure), but the effect is the same.

### Error Page
There is a custom OAuth error page now, but unless you modify the code or adjust some env variables you will hopefully not see it. If you are interested in doing this, let me know.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,10 +7,10 @@ OAUTH2_ADMIN_URL=http://hydra.auth.reaction.localhost:4445
 OAUTH2_AUTH_URL=http://localhost:4444/oauth2/auth
 OAUTH2_CLIENT_ID=example-storefront
 OAUTH2_CLIENT_SECRET=CHANGEME
+OAUTH2_PUBLIC_LOGOUT_URL=http://localhost:4444/oauth2/sessions/logout
 OAUTH2_HOST=hydra.auth.reaction.localhost
 OAUTH2_IDP_PUBLIC_CHANGE_PASSWORD_URL=http://localhost:4100/account/change-password?email=EMAIL&from=FROM
 OAUTH2_IDP_HOST_URL=http://identity.auth.reaction.localhost:4100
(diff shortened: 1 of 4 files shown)


willopez (approved):
Looks good, there is a minor issue with escaping `&` in the setup script that will be addressed separately.

PR #645, opened 2020-01-21
Update packages and solve high vulnerability

chrispotter:
Signed-off-by: Chris Potter <chris@reactioncommerce.com>

Impact: **minor**
Type: **chore**

## Issue
Updates dependencies and solves 1 high priority issue


HarisSpahijaPon:
Uhh, im not sure but this should not be able to work. Material UI needs a lot of refactoring in order to update it.

janus-reith:
@HarisSpahijaPon.

This is correct. Mostly it was changing theme.spacing.unit and the multiplications done with it to the spacing function, e.g. theme.spacing.unit * 2 => theme.spacing(2).

Also, the Grid spacing changed to be a multiple of the base unit, in our case 8, so sometthing like `<Grid spacing={24}> `became` <Grid spacing={3}>`

There also were some changes regarding SSR, but I can't exactly pinpoint what happended where, as I also upgraded to latest styled-components in the process. The final implementation got easier and jss could be removed, also pageContext doesn't need to passed down anymore.

chrispotter:
Thanks y'all, I'll have to re-evaluate with a less-broad-stroke to update only necessary packages!

PR #665, opened 2020-04-12
Storefront v2

### LINKED ISSUE #661: Redirect for reauth should not present login form
## Issue Description
Currently, when the storefront recieves a Status 401 from the backend over GraphQL, it will redirect to the `/signin` route. This will then redirect to the meteor backend to get a new authentication token. The assumption here seems to be, that the Meteor User still is logged in. If the User is logged out on the Backend, this will just present a login form. Skipping the auth redirect won't solve anything, as the page will just stay at its current point and display a Status 500 then.

[example-storefront/src/lib/apollo/initApollo.js](https://github.com/reactioncommerce/example-storefront/blob/ff29b3ec8bc72e9034056b90fd5a4940e3bbf348/src/lib/apollo/initApollo.js#L50)

The desired behaviour should be: If GraphQL gives a 401, get an actual indication if there is a logged in backend user.  If there is, do the redirect to get a new token. If not, ditch the current token and reload (Although I'd be unhappy with a real reload, maybe Apollo Client could just try to connect a second time )

## Steps to Reproduce
Please provide starting context, i.e. logged in as a user, configure a particular payment method.
1. Log in to the storefront
2. Wait until the token expires
2.5 Im not sure when actually the backend logout happens
3. Visit the storefront again, and see that it will redirect to the backend login form.

## Possible Solution
I'm unsure if that can be fixed in the backend on the route that also provides the silent reauth, or has to be done in the storefront.

We currently don't separate actual logins from silent reauths in the `/login`endpoint on the Backend:
https://github.com/reactioncommerce/reaction/blob/746f5acf8eb2a9e877d7ede93f46a595b8013348/imports/plugins/core/hydra-oauth/server/oauthEndpoints.js#L18

We could instead provide both a `/login` and `/refresh` (or `/reauth` as it isn't really a refresh in oauth terminology) endpoint. While the behaviour for both would be the same when `getLoginRequestRes.skip` is set, for the other case `/login` could still redirect to the login form as usual, while `/refresh` would redirect back to the storefront instead. I The storefront would have dropped the session cookie by then so it won't try the reauth again.

I think it would be even better if the unsuccessful redirect could be completely avoided, as it would happen quite often for logged in users. But the storefront would need to know beforehand if there is an authenticated backend user, and I'm not sure if that could work, as I think the Meteor Client bundle will always need to load once and grab the token from `localStorage` to which no other page has access. A hidden iframe solution could maybe work, that would only trigger a redirect if silent re-authentication is possible.

Im still need to figure out why exactly this issue occurs where the backend doesn't have valid authentication anymore while the frontend still thinks it is logged in.

### Versions
2.1.0

janus-reith:
Resolves #661
Impact: **breaking**
Type: **refactor**

## Issue
TBD

## Solution
TBD

## Testing
TBD


janus-reith:
Default email changed, guess that's why DCO failed - Will have to rebase.

PR #682, opened 2020-06-04
test: fix broken unit test

### LINKED ISSUE #681: Fix broken tests
Type: **critical**

**Describe the bug**
All unit tests are failing

mikemurray:
Resolves #681 
Impact: **critical**
Type: **test**

## Issue

tests are failing

## Solution

- add missing dependencies
- fixed tests themselves as necessary

## Breaking changes

none

## Testing
1. Unit tests should pass
1. App should run and be usable


mikemurray:
Fixed the `unit` and `dockerfile` issue. `eslint` issues are in another PR in the works

willopez (approved):
LGTM

janus-reith:
I wonder which of the tests rely on babel? I removed the customized babel config on purpouse as it is not required anymore. If tests really rely on it, maybe we could find a way so it is not picked up by nextjs?

Or do you want to include a custom babel config either way?

willopez:
@janus-reith there are only a few unit test that are running: 
```
 PASS  lib/utils/variantById.test.js (5.81 s)
 PASS  lib/utils/cartItemsConnectionToArray.test.js (5.922 s)
 PASS  lib/utils/pagination.test.js (6.188 s)
 PASS  lib/utils/relayConnectionToArray.test.js
 PASS  lib/utils/isProductLowQuantity.test.js
 PASS  lib/utils/isProductBestseller.test.js
 PASS  lib/utils/isProductOnSale.test.js
 PASS  lib/utils/priceByCurrencyCode.test.js
 PASS  lib/utils/calculateRemainderDue.test.js
```
10 to be exact, if we can get those to pass without adding the config in this PR, that would be great.

janus-reith:
@willopez Thanks, I also realized after taking another look which ones were meant.
@mikemurray also listed the remaining test locations in https://github.com/reactioncommerce/example-storefront/issues/683

I just looked at the nextjs with-jest example: https://github.com/vercel/next.js/tree/canary/examples/with-jest
There also is a babel config present in the root folder, which just uses `presets: "next/babel"`. I'm not an expert on the tests tooling, so maybe that's actually the way to go.
If we would just use that one, the issue @mikemurray described in https://github.com/reactioncommerce/example-storefront/pull/685 should also be gone I think? (But didn't try yet).

Issue #683, opened 2020-06-05
Fix & re-enable tests or remove unnecessary code and associated tests


mikemurray:
Type: **minor**

**Describe the bug**

Some tests, of the few 14, no longer work at all due to a massive update. Listed below are the places where the tests live that need to be revisited.

- `lib/tracking/utils` // Disabled in package.json jest config
- `tests/integration/ssr.test.js` // `.skip` added to suite
- `lib/utils` (Moved mock data into closer folder to fix missing imports)

### Loose directions
- Use React testing library https://github.com/testing-library/react-testing-library (get rid of enzyme, and other unnecessary testing dependencies)
- Remove or fix test locations above
- Remove unused utils
- Add unit tests for utils that are used
- Add tests for hooks and GQL
- Consider some tests for components. (think non brittle testing with getByRole and getByTest)
- Work towards a path of having E2E tests with cypress or something else.

reith-outgrow:
Maybe let's have a general overview of test we want to have in the storefront?
@mikemurray  Which one would you consider important?
Also, some options regarding the toolchain we could use changed the past years, e.g. we could decide between @testing-library/react, enzyme or react-test-renderer and I don't have a clear recommendation.

I'd like to see E2E tests introduced at some point for all the basic steps like adding to cart/checking out, most probably with cypress, and I'd like to keep them as broad in scope as possible so they would even continue to work with many storefront customizations, or would at least be easy to fix again.

Apart from that, it probably make sense to keep and maybe refactor the test for the util functions in lib/utils, I don't expect much things changing or breaking there. (Apart from the issues with babel test env as discussed)

Then it would probably also make sense to have tests available for the hooks and context Providers, as they account for much of the logic and probably will even more in some future release. (I could see them potentially becoming a separate library at some point, so I think there should be some focus on them.)

For the GraphQL Calls in there we could use `MockedProvider` which Apollo provides.
For that, I wonder what workflow we could maybe adopt to keep up with backend changes, e.g . instead of editing manually, to extract GraphQL API answers somehow or detect changes using Introspection somehow (not sure if feasible but it sounds great to not have to write most Mocks by hand).

We could have tests to detect major render changes, but preferably no Snapshots test, which would break soon when doing customizations and would eventually just be ignored by devs.

Regarding per Component testing, I'm somewhat uncertain as to what to reasonably test and to what extent, you will probably have a better overview here.

janus-reith:
[...]

PR #685, opened 2020-06-05
fix: runtime build issues

mikemurray:
## Issue

Cannot run app due to regenerator runtime issues

## Solution

Update babel config

## Breaking changes

none

## Testing
1. App should run and be usable


willopez (approved):
LGTM

janus-reith:
@mikemurray The babel config I made obsolete for the new version was added back in https://github.com/reactioncommerce/example-storefront/pull/682, introducing this error.
Instead of adjusting the babel config, I'd consider just removing it again.

I don't see us using specific features that would make it reasonable to introduce potential upstream issues by diverging from the nextjs defaults here.

willopez:
@janus-reith just saw your comment, after I approved it. I am open to removing it and simplifying the config. PR welcomed.

janus-reith:
@willopez Yes, I would prefer if we discuss in https://github.com/reactioncommerce/example-storefront/issues/683 or some seperate issue instead how to approach the tests cleanly, and which should be there in general.( I think @HarisSpahija also wanted to contribute to that.)

The previous storefront version also had a bunch of babel dependencies I'd rather not have included here again.

TBH my preference would be to revert https://github.com/reactioncommerce/example-storefront/pull/682 and not merge this one, and fix the tests instead.
Maybe a short term solution could be to have that babel config in a separate test-specific folder? Or any other way to scope it to the tests only, if it actually is necessary.

willopez:
Sounds good, lets discuss on #682

willopez:
This PR removes unnecessary babel config and some dependencies. A new PR will be created to remove snapshot test and related dependencies add react testing library and cypress. See for more context: https://github.com/reactioncommerce/example-storefront/issues/683#issuecomment-644266243

PR #686, opened 2020-06-16
Setup react-testing-library and some clean up

willopez:
Impact: **minor**
Type: **refactor**

## Issue
There were a couple of unnecessary dependencies related to testing.

## Solution
Removed unnecessary dependencies and added `react-testing-library` as the preferred testing library. Also, a sample test was adde for the `Footer` component. This is a stepping stone in the general testing direction described in https://github.com/reactioncommerce/example-storefront/issues/683

## Breaking changes
Any tests that used enzyme will no longer work.

## Testing
1. App start and run without issues
2. All tests should pass.


components/Footer/Footer.test.js
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+import { render } from "@testing-library/react";
+import Footer from "./Footer";
+
+test("Renders the footer", () => {
+  const { getByText } = render(<Footer />);
```
mikemurray: You should import `screen` then use `screen.getByText` instead. This is now the recommended way to use react-testing rather than de-structuring utils from the render func. Most of the utils you're used to using are available on `screen`.

```js
import { render, screen } from "@testing-library/react";

render(<Footer />);

screen.getByText(TEXT);
```

mikemurray (changes requested):

components/Footer/Footer.test.js
willopez: refactored.

mikemurray (approved):
👍

PR #690, opened 2020-06-18
Use CANONICAL_URL for sitemap route

### LINKED ISSUE #689: Sitemap Handler produces urls with undefined protocol
Type: minor

**Describe the bug**
The sitemap handler uses `req.protocol` to determine wether to use http or https for the links it returns:
https://github.com/reactioncommerce/example-storefront/blob/ddd0d069ae9f4498b98c8ca8acb79aa532bd3126/pages/api/sitemap.js#L12

`req.protocol` is an express feature not present in nextsjs api routes.
All it does is return `req.connection.encrypted ? "https" : "http"`

**To Reproduce**
1. Set up a Sitemap in the api
2. Check /sitemap.xml on your storefront, urls start with `undefined://`

**Expected behavior**
The urls in the sitemap contain the protocol that was used when requesting the sitemap.

**Additional context**
I tried to easily fix this by adding `req.connection.encrypted ? "https" : "http"` in the place of `req.protocol`, but the result on production deployments was that it always returns http - Could be, because the Reverse Proxy is handling SSL Termiantion, so nextjs is in fact not using https internally. This would mean that the same issue should've been present in older releases using express, but I currently have no production deployment for that at hand to verify it.
Maybe expess also does other thing to determine the requested url, and to my surprise I found no quick way in the next api routes to retrieve it properly.

*Solution*
I think the original approach of building the shop url by parsing `req.protocol` and `req.headers.host` separetely to get the url isn't really necessary anyway, and we could just resort to using `CANONICAL_URL` for this instead.

janus-reith:

Resolves #689
Impact: minor
Type: bugfix

## Issue
Sitemap routes currently show `undefined://` as `req.protocol` is not available in api routes.

## Solution
As detailed in #689, this will use `CANONICAL_URL` and pass it to the sitemap query instead of parsing the url from the incoming request.

## Breaking changes
None - I consider it unlikely that shops request sitemaps from a host name that differs from the one defined in `CANONICAL_URL`. If that is the case, these could adjust their sitemap api route individually.

## Testing
1. Make sure your api provides a sitemap
2. Navigate to http://localhost:4000/sitemap.xml and make sure it returns proper urls in the `loc` field.

### DIFF
--- a/pages/api/sitemap.js
+++ b/pages/api/sitemap.js
@@ -1,6 +1,9 @@
 import fetch from "isomorphic-unfetch";
 import appConfig from "../../config.js";
 
+let shopUrl = appConfig.CANONICAL_URL;
+if (shopUrl.endsWith("/")) shopUrl = shopUrl.slice(0, -1);
+
 /**
  * Generates the sitemap and returns it
  *
@@ -9,7 +12,6 @@ import appConfig from "../../config.js";
  * @returns {String} the the sitemap
  */
 export default async function generateSitemap(req, res) {
-  const shopUrl = `${req.protocol}://${req.headers.host}`;
   const handle = req.url.replace("/", "");
 
   const query = `


willopez (approved):
LGTM, thanks

balibebas:
Could this be patched onto `3.1.0` since the 4.0 release may be a ways out?

ghost:
Could this be backported and patched onto `3.1.0` since the 4.0 release may be a ways out?

PR #694, opened 2020-07-07
fix: Remove .env.prod file and its use as this causes issues with images built by CI

willopez:
Impact: **critical**

## Issue
Images built by CI with `env.prod` don't work correctly when deployed to our internal testing environment. It seems that vars in that env file override values provided to the container by its environment.  

## Solution
Remove the use of `env.prod` file and update README.

## Breaking changes
None

## Testing
1. Build production image locally and verify it build successfully
2. Deploy to test environment and verify it works as expected.


willopez:
not needed.

PR #696, opened 2020-07-08
Return Not found if either product or shop are missing

### LINKED ISSUE #695: Builds currently fail due to the page for the default product slug rendering the ProductDetail component
Type: **critical**

**Describe the bug**
Builds currently fail with `TypeError: Cannot read property 'media' of undefined` when the /[lang]/product route is built.

**To Reproduce**
1. Run a production build

**Expected behavior**
Builds don't fail. As the default slug "-" usually won't exist in the api during build, a simple "Not found" page is created.

janus-reith:

Resolves #695
Impact: **critical**
Type: **bugfix**

## Issue
This should fix building the pdp for the default slug "-" during build

## Solution
If there is no product, the further tree of components in the tree shouldn't render, and a "Not found" message is shown instead.
Instead of checking if both product and shop are missing, return that message each time one of both is missing.

## Testing
1. Run the production build
2. Observe that the build finished properly

### DIFF
--- a/pages/[lang]/product/[...slugOrId].js
+++ b/pages/[lang]/product/[...slugOrId].js
@@ -85,7 +85,7 @@ function ProductDetailPage({ addItemsToCart, product, isLoadingProduct, shop }) 
   }, [product, shop]);
 
   if (isLoadingProduct || router.isFallback) return <PageLoading />;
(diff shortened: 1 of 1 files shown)


willopez (approved):
LGTM and thanks!

PR #698, opened 2020-07-08
Revalidate index page, require instant revalidation if no shop is present during build

### LINKED ISSUE #697: The index page is not using revalidation, stays empty on production builds if no shop present during build
Type: **major**

**Describe the bug**
The index page fetches the shop data during build time, and passes its `_id` down to the clientside Query for catalog items.
As the shop is not present during the default CI build(and I suggest you to change that if possible), and no revalidation is set for that page. the shop id stays undefined and the query for catalog items is skipped.

**To Reproduce**
1. Run the production build
2. Visit the index page, no catalog items come up

**Expected behavior**
There are catalog items

janus-reith:

Resolves #697
Impact: **major**

## Issue
The shop is not present during the default CI build and no revalidation is set for that page. the shop id stays undefined and the query for catalog items is skipped.

## Solution
Set a revalidation time for that page, check if no shop is present and return fallback values that will require immedate revalidation.

## Testing
1. Run the production build
2. Visit the index page, no catalog items come up

## Additional context
As discussed with @willopez, we should somehow make sure that this is actually discouraged for production builds, as depoyments should not feature an initial render that is blank until the revalidated page is returned for subsequent requests.

### DIFF
--- a/pages/[lang]/index.js
+++ b/pages/[lang]/index.js
@@ -95,11 +95,27 @@ class ProductGridPage extends Component {
  * @returns {Object} the props
  */
 export async function getStaticProps({ params: { lang } }) {
+  const primaryShop = await fetchPrimaryShop(lang);
+  const translations = await fetchTranslations(lang, ["common"]);
+
+  if (!primaryShop) {
+    return {
+      props: {
+        shop: null,
+        ...translations
+      },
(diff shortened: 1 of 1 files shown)


willopez (approved):
LGTM

PR #700, opened 2020-07-10
Set _isMounted to true after component mounted

### LINKED ISSUE #699: infinite loader if the payment is rejected
Type: **minor**

**Describe the bug**
When you are placing and order on the checkout, using X payment gateway, If the payment is reject or fails the loader waiting for the order to place, never stops and we can't notifying the user that the payment failed.

**To Reproduce**
1. Add a product to the cart
2. Go to the checkout and fill step 1 and 2
3. Select your payment method and make sure it fails, for example entering a wrong credit card number if you use stripe
4. Place the order
5. See the infinite loader

**Expected behavior**
Stop the loader and display the alert with the error. CheckoutActions already do that but [this](https://github.com/reactioncommerce/example-storefront/blob/bad4cce75be188de7bcda4c894351e2db592dbe6/components/CheckoutActions/CheckoutActions.js#L219) conditional is not letting the component set the alert. For some reason the component is already unmounted and the component can't update the state

**Screenshots**
image

janus-reith:

Resolves #699
Impact: **minor**
Type: **bugfix**

## Issue
In https://github.com/reactioncommerce/example-storefront/commit/0ae583eb5f3f7a96270d968b536c770773a3cc38#diff-9b6019bce1be1df5431a56f4a139663dL77 I removed the whole `componentDidMount` part, not realizing that `this._isMounted = true;` was unrelated to the tracking events underneath.

## Solution
Add back a `componentDidMount` lifecycle function to set `_isMounted` to true in the first place.
Don't close #699 until confirmed that there is no separate issue present on older versions, where the component might actually not be mounted properly.

## Testing
See #699

### DIFF
--- a/components/CheckoutActions/CheckoutActions.js
+++ b/components/CheckoutActions/CheckoutActions.js
@@ -69,6 +69,10 @@ class CheckoutActions extends Component {
     }
   }
 
+  componentDidMount() {
+    this._isMounted = true;
+  }
+
   componentWillUnmount() {
     this._isMounted = false;
   }


manueldelreal (approved):
functionally it seems to be ok, I was able to checkout, and get an error message when using stripe and when disallowing guest checkout (to avoid IOU from completing)

willopez (approved):
LGTM

willopez:
Thanks @janus-reith!

PR #702, opened 2020-07-13
fix various bugs

willopez:
Impact: **minor**
Type: **bugfix**

## Issue
Fixes various small issues.

1. The `availablePaymentMethods` query should not executed if the shop id is not yet available
2. An error was being returned by an api route when the user was anonymous, it should not be an error, just a flag that says that the user is anonymous.
3. Add a not about a know issue with the anonymous cart query being executed unnecessarily, this is due to a know issue with Apollo Client: https://github.com/apollographql/react-apollo/issues/3492

## Testing
1. Start with a fresh installation
2. Observe and verify that the api route for /token is NOT returning an error on any view.
3. Observe and verify that in the checkout page, the `availablePaymentMethods` query dose not return an error.


ryan-veracitek:
We're seeing this bug "The availablePaymentMethods query should not executed if the shop id is not yet available" and the above fix does not appear to have made it to the code. In our storefront we cannot click checkout as storeid becomes null & language becomes German.

PR #705, opened 2020-07-15
fix: don't execute unnecessary requests for anonymous cart on initial load

### LINKED ISSUE #701: Unnecessary request for anonymous cart and account cart
Type: **minor**

**Describe the bug**
On a clean installation, request for the anonymous cart and account cart will be executed even though, they are explicitly directed not to by the `skip` param. This is a known issue with Apollo Client, [https://github.com/apollographql/react-apollo/issues/3492](https://github.com/apollographql/react-apollo/issues/3492). Revisit this after Apollo Client v3 is released and hopefully fixed. 

**To Reproduce**
1. Start with fresh installation
2. Load product grid page
3. Observe the following query being executed: 

```
anonymousCartByCartIdQuery
with params: 
"cartId":null,
"cartToken":null
```

**Expected behavior**
This query should not execute on first executed until a user clicks on the "add to cart" button. See, [https://github.com/reactioncommerce/example-storefront/blob/trunk/hooks/cart/useCart.js#L50](https://github.com/reactioncommerce/example-storefront/blob/trunk/hooks/cart/useCart.js#L50) for details.

 - OS: MacOS
 - Browser: Chrome
 - Version 83

willopez:
Resolves #701 
Impact: **minor**
Type: **bugfix**

## Issue
1. Unnecessary requests for anonymous cart are executed when they should not. The anonymous cart query should only executed after a user has added an item to their cart.
2. An error was being returned by an api route when the user was anonymous, it should not be an error, just a flag that says that the user is anonymous.

## Solution
1. Use `lazyQuery`s to control when cart requests are executed.
2. Return a boolean flag `anonymous` to denote when a user is anonymous

## Testing
1. Load product grid and verify no `anonymousCartByCartIdQuery` request are executed.
2. Add and remove items to and from the cart and verify they are add/removed as expected.
3. Complete a checkout flow and place an order
4. Verify there are no error for the `api/token` route

### DIFF
--- a/hooks/cart/useCart.js
+++ b/hooks/cart/useCart.js
@@ -51,7 +51,7 @@ export default function useCart() {
 
   const [
     fetchAnonymousCart,
-    { data: cartDataAnonymous, called: anonymousCartQueryCalled, refetch: refetchCartAnonymous }
+    { data: cartDataAnonymous, called: anonymousCartQueryCalled, refetch: refetchAnonymousCart }
   ] = useLazyQuery(anonymousCartByCartIdQuery, {
     variables: {
       cartId: cartStore.anonymousCartId,
@@ -71,7 +71,7 @@ export default function useCart() {
       refetchAccountCart();
     }
     if (!shouldSkipAnonymousCartByCartIdQuery && anonymousCartQueryCalled) {
-      refetchCartAnonymous();
+      refetchAnonymousCart();
     }
   }, [viewer, refetchAccountCart]);
 
@@ -126,7 +126,7 @@ export default function useCart() {
       if (accountCartId) {
         refetchAccountCart();
       } else if (anonymousCartToken) {
-        refetchCartAnonymous();
+        refetchAnonymousCart();
       }
     }
   });
--- a/hooks/cart/useCart.js
+++ b/hooks/cart/useCart.js
@@ -37,7 +37,10 @@ export default function useCart() {
   const shouldSkipAccountCartByAccountIdQuery = Boolean(!accountId || cartStore.hasAnonymousCartCredentials || isLoadingViewer || !shop || !shop._id);
   const shouldSkipAnonymousCartByCartIdQuery = Boolean(accountId || isLoadingViewer || !cartStore.anonymousCartId || !cartStore.anonymousCartToken);
 
-  const [fetchAccountCart, { loading: isLoading, called: accountCartQueryCalled,  data: cartData, fetchMore, refetch: refetchAccountCart }] = useLazyQuery(accountCartByAccountIdQuery, {
+  const [
+    fetchAccountCart,
+    { loading: isLoading, called: accountCartQueryCalled, data: cartData, fetchMore, refetch: refetchAccountCart }
+  ] = useLazyQuery(accountCartByAccountIdQuery, {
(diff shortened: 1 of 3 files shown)


manueldelreal (approved):
LGTM! are we including this in the pending release branch?

hooks/cart/useCart.js
```diff
@@ -37,32 +37,43 @@ export default function useCart() {
...
     },
     pollInterval: shouldSkipAccountCartByAccountIdQuery ? 0 : 10000
   });
 
-  const { data: cartDataAnonymous, refetch: refetchCartAnonymous } = useQuery(anonymousCartByCartIdQuery, {
-    skip: shouldSkipAnonymousCartByCartIdQuery,
+
+  const [
+    fetchAnonymousCart,
+    { data: cartDataAnonymous, called: anonymousCartQueryCalled, refetch: refetchCartAnonymous }
```
janus-reith: I think It was a good idea to rename `refetchCart` to `refetchAccountCart` to avoid confusion, but I would rename `refetchCartAnonymous` to `refetchAnonymousCart` accordingly.
```suggestion
    { data: cartDataAnonymous, called: anonymousCartQueryCalled, refetch: refetchAnonymousCart }
```

janus-reith (changes requested):
I think using Lazy Query currently is the best solution.
Although the related errors in the Apollo-client repo were closed with the release of 3.0.1, multiple people still observed the described error.
I also tried upgrading apollo-client to the latest v3.0.2 and as that didnt work, also upgraded the remaining apollo packages that we're using to their latest version, also without any success.

After looking at your changes I checked out the PR locally and can confirm that the error is fixed.
Tried out anonymous carts, account carts and reconciliation, all seem to work fine.

The only minor change I'd like to suggest is the wording for `refetchAccountCart` / `refetchCartAnonymous`.

hooks/cart/useCart.js
janus-reith: As described above
```suggestion
    { data: cartDataAnonymous, called: anonymousCartQueryCalled, refetch: refetchAnonymousCart }
```

hooks/cart/useCart.js
```diff
@@ -37,32 +37,43 @@ export default function useCart() {
...
+
   useEffect(() => {
-    if (!shouldSkipAccountCartByAccountIdQuery) {
-      refetchCart();
+    if (!shouldSkipAccountCartByAccountIdQuery && accountCartQueryCalled) {
+      refetchAccountCart();
     }
-    if (!shouldSkipAnonymousCartByCartIdQuery) {
+    if (!shouldSkipAnonymousCartByCartIdQuery && anonymousCartQueryCalled) {
       refetchCartAnonymous();
```
janus-reith: As described above
```suggestion
      refetchAnonymousCart();
```

willopez:
@manueldelreal Yes, we will include this PR in the upcoming release.

willopez:
@janus-reith Thanks for the suggestions and testing your suggestion has been implemented.

janus-reith:
@willopez Looks like the upcoming v3.1.2 might fix this, so we could remove that workaround again: https://github.com/apollographql/apollo-client/pull/6752

PR #723, opened 2020-08-30
perf: stop stripe scriptloader when stripe is disabled

### LINKED ISSUE #722: Disable stripe script loader when Stripe is not a defined payment method
Type: **minor**

**Describe the bug**
Stripe script loader loads even when stripe is disabled. The script loader kicks off three http requests which lead to blocking in the run loop even when Stripe is explicitly disabled as a payment method.

**To Reproduce**
1. Disable the Stripe payment method hard-coded in the storefront `paymentMethods` array.
2. Reload the page.
3. Observe call to `js.stripe.com` script loader and related requests.
4. See guidance from Google included in Screenshots.

**Expected behavior**

**Screenshots**

Screen Shot 2020-08-31 at 04 17 01
Stripe blocks the main thread for almost a third of a second even when Stripe payment method is disabled.

 - OS: macOS Big Sur beta
 - Browser: Vivaldi
 - Version: 3.2.1967.45 (Stable channel) (64-bit)

**Additional context**
The payment methods hard-coded in the app are good as they prevent the need to wait for an API call to determine if Stripe should be available, so let's take advantage of them to give performance back to the users who cannot use Stripe (and are more likely to use Reaction Commerce as a result).

ghost:
fixes: #722
Impact: **minor**
Type: **perf**

## Issue
See backing issue.

## Solution
checks defined payment methods for stripe_card before loading

## Breaking changes
n/a

## Testing
1. Disable stripe payment method hard-coded in `paymentMethods.js`
2. Do page reload with DevTools open.
3. Verify no call is made to `js.stripe.com`
4. Enable the IOU checkout method.
4. Add a product to the cart.
4. Start checkout and proceed to payment.
4. Verify Stripe does not appear as a possible payment method.

### DIFF
--- a/pages/_document.js
+++ b/pages/_document.js
@@ -4,7 +4,8 @@ import { ServerStyleSheet as StyledComponentSheets } from "styled-components";
 import { ServerStyleSheets as MaterialUiServerStyleSheets } from "@material-ui/styles";
 import favicons from "custom/favicons";
 import theme from "custom/reactionTheme";
-import analyticsProviders from "../custom/analytics";
+import definedPaymentMethods from "custom/paymentMethods";
+import analyticsProviders from "custom/analytics";
 
 /**
  * For details about the styled-components SSR code in this file, see https://www.styled-components.com/docs/advanced#nextjs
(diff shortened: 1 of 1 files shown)


janus-reith (approved):
Looks fine to me.

BTW, a newly implemented version of the stripe payments could help getting rid of this altogether by just using `useStripe` from `@stripe/react-stripe-js` contained inside the Stripe payment component(which then only loads if that renders).
We would also need to get rid of the current hard requirement coming from the component-library, which throws an error if the deprecated `react-stripe-elements` is not installed in the storefront, even if Stripe is not in use anywhere.

janus-reith (approved):

janus-reith (approved):
Hm, seems like I lack the permissions to make that pass the check, maybe @kieckhafer could have a look?

kieckhafer (approved):

ghost:
Thanks for the merge folks. If I'm able to use Stripe in the future (currently they don't allow sales of M. speciosa) I'll consider the benefit in helping with a lib upgrade to take better advantage of the dev affordances mentioned by @janus-reith.

PR #731, opened 2020-09-30
feat: add tracker pages, events and identify

### LINKED ISSUE #676: Remove or refactor segment analytics
Type: refactor

**Describe the refactoring**
Remove or replace `custom/analytics/provider.example.js` and `custom/analytics/segment.js` depending on how we implement this. For additional context see: https://github.com/reactioncommerce/example-storefront/pull/667#issuecomment-616207068

jhonsfran:
Segment and google analytics support

Resolves #676 
Impact: **minor**
Type: **feature**

## Testing
1. Add a Segment key and a google track-id to the .env file
2. Start the project
3. See in the segment dashboard if the events are reported


jhonsfran:
Hi @janus-reith let me know what do you think of those changes.

janus-reith:
Thanks @jhonsfran1165 , I will have a look soon!

hooks/viewer/useViewer.js
```diff
@@ -25,7 +28,27 @@ export default function useViewer() {
   }, [accessToken, viewer]);
 
   useEffect(() => {
-    if (viewer) setAccount(viewer);
+    if (viewer) {
+      setAccount(viewer);
+
+      // identify the user
+      const isAuthenticated = viewer?._id;
+      const isTracked = localStorage.getItem("isTracked");
```
janus-reith: Is this really necessary?

If `viewer` changes, I'd guess it's okay to send another identify call / prefer risking duplicate calls over outdated data - If any provider is not designed to deal with this for some reason, I guess some identify logic could be applied on the provider level to hash values and avoid duplicates.

Right now, if I log out and next log in in with another account, the new user would not be identified with this in place if I get it correctly.

lib/analytics/index.js
```diff
@@ -0,0 +1,22 @@
...
+      writeKey: process.env.SEGMENT_ANALYTICS_WRITE_KEY
+    }),
+    googleAnalytics({
+      trackingId: process.env.GOOGLE_ANALYTICS_ID
+    })
+  ]
+});
+
+/** Disable plugins that it's not configured properly */
+if (process.env.GOOGLE_ANALYTICS_ID === "") analytics.plugins.disable(["google-analytics"]);
```
janus-reith: I'm wondering, instead of a separate call to disable plugins in a separate call after loading them, couldn't we just conditionally enable them based on the env value right at the definition of the `plugins` array?

pages/_app.js
```diff
@@ -1,25 +1,51 @@
...
+        path: url,
+        query,
+        shopId: shopDecoded?.id
+      });
+    };
+    router.events.on("routeChangeComplete", handleRouteChange);
+    return () => {
+      router.events.off("routeChangeComplete", handleRouteChange);
+    };
+  }, [router.events]);
```
janus-reith: My preference would be to avoid adding more complexity right to this file, instead creating a separate hook for this, which could also include the logic you added in `lib/analytics/index.js` to initialize Analtics and return that instance.

So in _app.js you would only have something like e.g.  `const analytics = useInitAnalytics()` and pass that instance to `AnalyticsProvider` as you do right now.
Inside of that hook, you could declare that effect as part of the lifecycle, but still have the initialization of the analytics module done global to the file.

Some other things regarding the the call to `analytics.page`:

- I believe this will be executed on the server side during page creation aswell and is also supportedby the library, but not what we want, so we should probably check for `typeof window !== "undefined"` before and only execute that conditionally (no need to return it as done above).

- I could observe some issues with the nextjs routing event. as that somehow fired before the changed url was propagated properly. I don't recall the exact details, but the problem was there for quite some time, and still occured for the latest version I tested with around 4 weeks ago.

So I ended up with something like:
```js
if (typeof window !== "undefined") {
        setTimeout(() => {
          analytics.page({
            path: url,
          });
        }, 1000);
}
```

- I guess `previewUrl` is set to avoid duplicates, but A. I find the name a bit misleading, B. I Wonder if that is necessary as I wouldn't expect the event to fire multiple times for no reason, but if there indeed is a route transition to the same route, I see no issue with also tracking that.

- I believe `isFallback`, `query` and `shopId` are all non-default values and wonder if they are used by the included providers. While it is fine in general to pass custom pagedata here, did you check if the Providers actually parse it? I recall at least the Google Analytics plugin to provide some logic to pass them as custom "dimensions", so maybe it works fine.
Either way, I would prefer to not decode the shop Id here and pass it as-is, in the end the base64-encoded one is the "real" client-facing ID we have.

.env.example
```diff
@@ -19,3 +19,4 @@ SEGMENT_ANALYTICS_WRITE_KEY=ENTER_KEY_HERE
 SESSION_MAX_AGE_MS=2592000000
 SESSION_SECRET=CHANGEME
 STRIPE_PUBLIC_API_KEY=ENTER_STRIPE_PUBLIC_KEY_HERE
+GOOGLE_ANALYTICS_ID=ENTER_YOUR_GOOGLE_ANALYTICS_ID
```
janus-reith: Should we include Google Analytics by default?
I'm fine with it, but I'd like someone else from the core team to answer that.

hocs/withTracking.js
```diff
@@ -0,0 +1,20 @@
...
+    // Set the correct displayName in development
+  if (process.env.NODE_ENV !== "production") {
+    const displayName = PageComponent.displayName || PageComponent.name || "Component";
+    WithInjectedStores.displayName = `withInjectedStores(${displayName})`;
+  }
+
+  return WithInjectedStores;
+};
+
+export default withTracking;
```
janus-reith: I guess this was added as a workaround for now to avoid converting `ProductDetail` to a functional component?
`WithInjectedStores` sounds a bit misleading, as essentially, none of the global stores/context is added here.

hooks/analytics/useTrackingEvents.js
```diff
@@ -0,0 +1,22 @@
...
+import trackProduct from "lib/tracking/trackProduct";
+
+
+/**
+ * tracking events visits
+ *
+ * @param {String} data - the props of the page
+ * @returns {Object} useTrackingEvents
+ */
+export default function useTrackingEvents() {
```
janus-reith: Regarding the bigger picture, I'm a bit unclear on that one.
This is called `useTrackingEvents`, but it seems to cover no shared logic relevant for tracking the various events other than the `trackProduct` event?

Also, if we end up adding more of them here and maybe rename `callTracking`, the hook would get a bit bloated and declare several functions during render for each components somewhere making use one of the tracking events.

Maybe instead, call this one something like `useTrackProducts` which returns some `trackProduct` function?
To then give that hook some more purpose, import the constant `PRODUCT_VIEWED` right here instead of passing it as part of the payload from the `ProductDetail` component into the function.

Optionally, also include the functionality imported from `lib/tracking/trackProduct` right here as AFAIK we have no other usecase for that, but that's just an idea.

I don't want to be too picky about this, but I'd really want the result of this to be less cubersome and easier to understand than the previous implementation.

lib/tracking/trackProduct.js
```diff
@@ -1,43 +1,37 @@
...
-        action,
-        ...getProductTrackingData(product)
+        ...data,
+        ...getVariantTrackingData({
+          product,
+          router,
+          variant, // Object representing a variant. (Required)
+          optionId // Selected option of the provided variant, if available. (Optional)
+        }),
+        allDataProduct: product // Full product document for additional data. (Optional)
```
janus-reith: I'd imagine this to possibly get quite a big payload - is this picked up anywhere?

hooks/viewer/useViewer.js
```diff
@@ -25,7 +28,27 @@ export default function useViewer() {
   }, [accessToken, viewer]);
 
   useEffect(() => {
-    if (viewer) setAccount(viewer);
+    if (viewer) {
+      setAccount(viewer);
+
+      // identify the user
+      const isAuthenticated = viewer?._id;
```
ghost, line 40: Thinking out loud here. From a platform consistency perspective does it make sense to introduce optional chaining into the codebase right now when [it requires](https://node.green/#ES2020-features-optional-chaining-operator-----) Node 14.5+ to use? The dev docker image is currently using Node v12, so I'm curious if there are any environment changes required or if recent versions of Next.js just handle it like magic.

lib/tracking/trackProduct.js
```diff
@@ -1,43 +1,37 @@
...
+    // Add variant data if available
+    if (variant) {
       data = {
-        action,
-        ...getProductTrackingData(product)
+        ...data,
+        ...getVariantTrackingData({
+          product,
+          router,
+          variant, // Object representing a variant. (Required)
```
ghost: Is this covered in JSDoc somewhere so the editor will pull it up automatically using IntelliSense versus relying on comments?

.env.example
ghost: I'd prefer if GA was left out and, instead, stuck with Segment—which can report to GA anyway. GA is good for websites but not suitable IMO for web apps such as an e-commerce storefront. If the idea is to have multiple providers I'd prefer to have to add GA the same way [users have to add](https://docs.reactioncommerce.com/docs/3.0.0/environment-variables) `MAIL_URL` today.

hooks/viewer/useViewer.js
janus-reith: Yes @balibebas, Nextjs has optional chaining built-in since the beginning of this year, since some 9.x version.
The code is transpiled anyway, the docker node image doesn't play a big role here, some even older version would probably work aswell, official nextjs requirement is 10.13.

I like using optional chaining in more cases where it is possible, to reduce some verbosity, especially regarding data-fetching.

lib/tracking/utils/getProductTrackingData.js
```diff
@@ -1,16 +1,19 @@
...
 /**
  * Transform a product object into a partial representation of the Segment product schema.
  * Combine with `getVariantTrackingData(varaint)` to get the full definition
  * @name getProductTrackingData
  * @param {Object} product Project object
+ * @param {Object} router nextJs router
  * @returns {Object} Data for tracking
  */
-export default function getProductTrackingData(product) {
+export default function getProductTrackingData(product, router) {
```
janus-reith: I'd prefer if we could find a way to avoid passing the router here.

hooks/analytics/useTrackingEvents.js
jhonsfran: I agree, What do you think if I pass the productTrack function as a parameter from ProductDetail? Or maybe you're right and I can create a hook to each event. But the problem I see there is that I'm using a HOC to call that Hook. And in the ProductDetail I have to track other events like PRODUCT_ADDED.

hocs/withTracking.js
jhonsfran: I just changed this.

.env.example
jhonsfran: Got it

hooks/viewer/useViewer.js
jhonsfran: thanks I was not sure in that part if it was a good idea to allow duplicate calls

lib/tracking/trackProduct.js
jhonsfran: I didn't get the question? would you explain to me a bit more?

pages/_app.js
jhonsfran: thanks a lot! 🙌

jhonsfran:
@janus-reith thanks for all your suggestions, I learned a lot. When you have time let me know what do you think of my latest changes.

lib/tracking/trackProduct.js
janus-reith: What I mean is, if `allDataProduct` is sent alongside the event.
The full product could potentially be quite big, too big to send it with each event.
`trackProduct` would currently pick some relevant fields only and return only that data.
So I was wondering if the tracking providers currently pick up the content of `allDataProduct` (which would be bad) or if it is filtered out somewhere in the process, I didn‘t look that up yet.
If there is no other usecase of it, I‘d just want to remove it.

jhonsfran: Got it

hooks/analytics/useTrackingEvents.js
jhonsfran: is that a good idea or instead of doing that I can create a custom hook to each event (exactly as you said) and transform ProductDetail to a functional component?

janus-reith: My general preference on this is, that we should keep as much of the non ui-specific logic out of the component as possible, and put that behind some hook - If you don‘t want to convert a class components to functional to be able to use the hook, I‘m fine with some HOC wrapper using the hook like you did here for the pdp component already.

Regarding the `useTrackingEvents` hook, I think it‘s fine to keep it if it provides some general logic that can be used across multiple types of tracking events, but then I would make the specific event hook, e.g. useProductTrackingEvent, make use of that hook.

Does that make sense to you the way I put it? :)

ghost:
Small nit. I'd challenge this as a `minor` change because it affects how product owners and other stakeholders will perceive the data flowing into their storefronts and decisions made in a business are only as good as the data used to make them.

hooks/analytics/useTrackingEvents.js
jhonsfran: Hi, yes cristal clear :D 

I decided to refactor the ProductDetail component to a Functional component.

jhonsfran:
hi @balibebas what do you think is the impact of this? major?

ghost:
> hi @balibebas what do you think is the impact of this? major?

Minor if backwards compatibility with current Analytics events is achieved. Breaking/major if downstream users must update their Analytics reports and software due to changes in reporting and something which should be highlighted in a 4.0 migration guide. If event parity cannot be achieved the technical challenges in doing so will be the first thing consumers will want to know in order to determine if they need to customize so they don't, for example, break their businesses KPIs already coming through Segment (as I am, using Amplitude, for example).

ghost:
I'm dropping in a dashboard example I generate using the existing data flowing through on the `v3.1.0` release (which was automatic and I absolutely love it) as an example of how existing metrics might be being used to help highlight how variance in the way events are used could ripple through many different reports used by a business.

Screen Shot 2020-10-07 at 1 30 33 PM

jhonsfran:
> > hi @balibebas what do you think is the impact of this? major?
> 
> Minor if backwards compatibility with current Analytics events is achieved. Breaking/major if downstream users must update their Analytics reports and software due to changes in reporting and something which should be highlighted in a 4.0 migration guide. If event parity cannot be achieved the technical challenges in doing so will be the first thing consumers will want to know in order to determine if they need to customize so they don't, for example, break their businesses KPIs already coming through Segment (as I am, using Amplitude, for example).

I agree. So I think this is minor because I'm working on the same events that version 3.1.0 is tracking. Also the same structure of the data. It's a matter of time to track the missing events. But at the end segment will receive the same data.

jhonsfran:
@janus-reith @balibebas  could you give me some feedback about the latest changes? If all go well I'll work on the other events.

hooks/products/useProductDetail.js
```diff
@@ -0,0 +1,125 @@
...
+    product
+  } = props;
+
+  const selectVariant = (variant, optionId) => {
+    // Select the variant, and if it has options, the first option
+    const variantId = variant._id;
+    let selectOptionId = optionId;
+    if (!selectOptionId && variant.options && variant.options.length) {
+      selectOptionId = variant.options[0]._id;
+    }
```
ghost: Could this be simplified in a manner similar to:

```
const selectedVariantId = variant._id;
const firstOption = [variant.options];
const variantOrOptionId = optionId || firstOption?._id || selectedVariantId
```

Then used to remove the boolean logic from the `url` assignment?

ghost:
> @janus-reith @balibebas could you give me some feedback about the latest changes? If all go well I'll work on the other events.

I took a quick look at the last three commits. I'm not checked out on the branch—I prefer to test code I'm reviewing but I'm short on time—though I did spot one potential simplification. Other than that I think more JSDoc would be helpful since we're not yet using TypeScript though perhaps that (the JSDoc) is going to be added later?

package.json
```diff
@@ -34,6 +34,8 @@
     }
   },
   "dependencies": {
+    "@analytics/google-analytics": "^0.5.2",
```
ghost, line 37: Not sure why we'd want to pull in GA. Doing so will encourage people to avoid using a CDP (Customer Data Processor) like Segment, which is what most businesses will want anyway. On top of that Google Analytics is built more for Webpages and not Apps like this IMO. If users really want this they could add it on their own. Otherwise we can pull it and (a) still use GA because (b) users can customize if they like by adding this dependency.

EDIT: Here's the [current doc](https://github.com/reactioncommerce/example-storefront/blob/7fd2571f2f15c7e86ea06c4e7d79a7596d0c9fa1/docs/tracking-events.md) for tracking. It explains users can add their own tracking provider if they're not using Segment. I believe this needs to be updated as decorators (my personal preference) have been removed in favor of an HOC-wrapping approach near individual file exports.

hooks/products/useProductDetail.js
janus-reith: Thanks for the work you put into refactoring this @jhonsfran1165 !
I do agree that instead of all the separate calls to determine the variants in several places, we could just determine and memoize them at the top of the component and overall simplify this.

However I think that doesn't necessarily needs to be part of this PR, as this is complexity which also was there before.

package.json
janus-reith: I don't really agree. Neither to I think it should be much of a concern for us wether we should encourage to use segment or not, nor do I think that Google Analytics is not suited for this kind of ecommerce platform, I see no reason to differentiate between app or Webpages here.

I'm fine with keeping it, as long as it doesn't load per default if no `GOOGLE_ANALYTICS_ID ` was set, which does not seem to be the case.

yarn.lock
```diff
@@ -2146,6 +2187,36 @@ alphanum-sort@^1.0.0:
   resolved "https://registry.yarnpkg.com/alphanum-sort/-/alphanum-sort-1.0.2.tgz#97a1119649b211ad33691d9f9f486a8ec9fbe0a3"
   integrity sha1-l6ERlkmyEa0zaR2fn0hqjsn74KM=
 
+analytics-node@^3.4.0-beta.2:
```
janus-reith, line 2190: Need to check that this doesn't bundle anything unrelated we don't want on the client

docker-compose.dev.yml
```diff
@@ -8,7 +8,7 @@
 # To go back to running the published image:
 #  rm docker-compose.override.yml
 
-version: '3.4'
+version: "3.4"
```
janus-reith: @jhonsfran1165 Could You remove this?
Main docker-compose still uses single quotes, we should keep this consistent
```suggestion
version: '3.4'
```

hooks/analytics/useInitAnalytics.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Initialize the analytic provider
+ * @returns {Object} useInitAnalytics
+ */
+export default function useInitAnalytics() {
+  const router = useRouter();
+
+  useEffect(() => {
+    const handleRouteChange = (url) => {
```
janus-reith, line 13: Need to clarify: Does this track the first entry page aswell when this effect runs, or only subsequent route changes?
Might need to do a separate `analytics.page()` call at the beginning of the effect if that's the case.

hooks/useWidth.js
```diff
@@ -0,0 +1,21 @@
...
+import useMediaQuery from "@material-ui/core/useMediaQuery";
+
+/**
+ * Be careful using this hook. It only works because the number of
+ * breakpoints in theme is static. It will break once you change the number of
+ * breakpoints. See https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level
+ *
+ * @returns {Object} breakpoint of the windows
+ */
+export default function useWidth() {
```
janus-reith, line 11: I also saw this hook example in the MUI documentation to recrate the original `widthWith` hebaviour , I think it's fine to add it, there might be other usecases for it aswell.

janus-reith (changes requested):
Thanks for your on this @jhonsfran1165 and sorry for the delay, overall this looks fine to me, just maybe you could remove the altered quotes on `docker-compose.dev.yml` ?

Just a few things we need to check:
- [ ] Sill need to test around to ensure nothing breaks
- [ ] Need to ensure that `analytics.page()` actually fires on first page load, by just looking at the code I'm not sure about that
- [ ] Check bundle size, ensure that none of the dependencies related to analytics-node bundle unnecessary stuff.

I think it's fine to be merged then.

Two additional things that could be adjusted but are not a blocker for me:
- I'm not that happy with the url being passed between functions, I wonder if we could just make use of `useRouter` somewhere at the end of that where it's used.
- It's great that you refactored the PDP, I think we could also simplify some of the repetitive logic to determine the selected variants in there.

jhonsfran:
@janus-reith @balibebas thanks for the feedback guys. I've been working on this for a while. Hopefully, at the end of the week, I'll push it.

yarn.lock
jhonsfran: it seems good to me https://github.com/segmentio/analytics-node#readme

janus-reith: `analytics-node` is a dependency of `@analytics/segment`, which exports a node and a browser version, depending on `process.browser` being truthy. the node version uses `analytics-node`.

We are importing `@analytics/segment` global to a file imported in _app.js, so I was curious how this would be bundled and quickly made a little test sandbox, once building a mosty blank page, once importing `@analytics/segment`.
The JS bundle for that page was only increased by 0.9kb, indicating that only the small browser-related portion is bundled and not `analytics-node` and its range of dependencies like `axios`.

So yes indeed, everything seems good here :)

jhonsfran: Thanks! I appreciate the explanation. It's a good way to test it 😅

hooks/analytics/useInitAnalytics.js
jhonsfran: I did a couple of tests and it works for the first loading.

jhonsfran:
Hi @janus-reith I think it's done. There are some TODOS but I think we can work on them with a couple of changes in the API. Let me know what you think.

lib/tracking/utils/getCartItemTrackingData.js
```diff
@@ -1,39 +1,31 @@
...
     name: variant.title,
     brand: variant.productVendor,
     variant: variant.variantTitle,
     price: variant.price.amount,
     quantity: variant.quantity,
+    // TODO: get somehow this data - imageURLs is empty
     image_url: variant.imageURLs && variant.imageURLs.original, // eslint-disable-line camelcase
-    url
+    // TODO: get somehow this data if not possible keep only the product slug
+    url: `/${locale}/${variant.productSlug}`
```
jhonsfran, line 29: @janus-reith what do you think it's the best way to get this data?

lib/tracking/utils/getCartItemTrackingData.js
```diff
@@ -1,39 +1,31 @@
...
-
   let cartId = {};
+
   if (variant.cart_id) {
     cartId = { cart_id: variant.cart_id }; // eslint-disable-line camelcase
   }
 
   return {
     ...cartId,
-    product_id: variant.productConfiguration.productVariantId, // eslint-disable-line camelcase
```
jhonsfran, line 28: Here we were using the ProductVariantId but I think the right value is productId

Akarshit:
Closing right now. Reopen if this still needs to be merged.

PR #740, opened 2020-12-04
Circle chart

eastern1984:
1. Create a chart component based on the attached design. It should accept following properties

- label - string value ( in attachment it has value  Facebook ) 
- color - string value in format of hexadecimal color  ( it is used for adding color to a chart )
- value - number value between 0 and 1 ( in attachment it has value 0.67 or 67% )

2. Add that chart component on a homepage

- Label of that chart should be name of primary shop retrieved from GraphQL API
- Value of the chart can be any hardcoded value between 0 and 1
- color can be any hardcoded color value


eastern1984:
wrong repository

PR #745, opened 2021-01-27
add package-link file to storefront

kieckhafer:
Impact: **minor**
Type: **chore**

Adds `package-link` and `package-unlink`, similar to we have in the `api` project, to allow for dev linking of outside npm packages.

We need to update the base image to use `12.14.1-v3` instead of `12.14.1-v1` as this version installs `yalc` inside the container, which is required for this linking.

TESTING:
- clone this branch
- clone any plugin (component library in this example
- `bin/package-link @reactioncommerce/components ../reaction-component-library/package`
- `bin/package-unlink @reactioncommerce/components`


mikemurray:
@kieckhafer We should add the `package-update` script as well. It's probably less used, but still pretty helpful to refresh the already installed plugins quickly.

kieckhafer:
@mikemurray done.

bin/package-update
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
```
mikemurray, line 1: I this file needs execute permissions.

mikemurray:
Not sure why the push is failing.

mikemurray (approved):
👍

Issue #718, opened 2020-07-31
Build is dependent on a committed ENV file, and doesn't obey system ENV vars for GQL url


mikemurray:
Type: **critical**

**This is a blocker for releases**

**Describe the bug**

The `BUILD_GRAPHQL_URL` and/or EXTERNAL_GRAPHQL_URL env var is being baked into the JS bundle, likely due to the `Next.js` / `webpack` / static build. This bakes unchangeable configuration into the app, which leads to the built docker image on docker hub, being useless if you aren't using the default variables.

Part of the solution for enabling the `Next.js` static build was maintaining a `.env.prod` to supply those variables at build time for the production build. This env file is an anti-pattern for the following reasons.

1. Promotes a dangerous practice of committing environment vars to a repo in a plain text file.
1. Ties configuration to dev/prod when the build should have not care which system it's building for. The built app should run the same if you point it to your dev or production API.

ENV vars should always come from the system, and never be in a file that is committed to source control. The `.env` and `.env.example` files are a convenience local development only.

Because of the `.env.prod` file and the `Next.js` static build process, we're not able to use our generic Docker image published to Docker hub to run in our test environments, and generic users cannot use it as is without rebuilding.

**Expected behavior**

- Do not require a static build in `Next.js`. This might be related to https://github.com/reactioncommerce/example-storefront/issues/716
- Remove `.env.prod` and make the build not dependent on that file. It should get its vars from the environment, whether you put them there in a prefix to the command, or `cat` them into the env, or set them in CI.
- Fix the issue where the GraphQL urls are being baked into the built JS bundle. 
  - Maybe proxy GraphQL requests through a `Next.js` API route so we can take advantage of system environment vars
  - Somehow make sure config doesn't get bundled

janus-reith:
I think this should go into two separate issues, as the fact pages are statically generated at build time won't compromise security of environment vars. We don't necessarily need to provide env vars from file with static generation, although those used inside the client bundle as defined in [next.config.js](https://github.com/reactioncommerce/example-storefront/blob/f8dbd9e6a4800b3b3d87d193bfae88f989b35b90/next.config.js#L5) would be inlined at build time and not change until maybe the page is revalidated.

Regarding how .env vars could be handled, I brought that up here: https://github.com/reactioncommerce/example-storefront/issues/691#issuecomment-652568441 and there: https://github.com/reactioncommerce/example-storefront/pull/667#issuecomment-631460886.
[...]

PR #746, opened 2021-01-29
Don't get shop to pre-render page

### LINKED ISSUE #742: Checkout flow not functioning as expected when running in local production mode
Type: **major**

**Describe the bug**
When running the app locally in production mode the example-storefront checkout flow does not function as expected. When the user has placed items in their cart and clicks checkout, the items are removed from the cart preventing the user from checking out.  Additionally the language switches to German (this behavior is not consistent). However, hovering over the cart prior to clicking it shows the items in the cart as expected. After exiting the cart page, the on hover behavior of the cart once again shows the correct orders. This issue is only seen in production mode run locally.

**To Reproduce**
1. Start the app locally in production mode (`make`).
2. Ensure that you have a payment method enabled and have a shipping method listed in the admin portal.
2. Log into the example storefront.
3. Add items to cart if your cart is empty. You should be able to hover over the cart icon and see the popup appear as expected with the items in your cart.
4. Click on the cart icon.
5. You should see "Your shopping cart is empty" and the language will now be German. On hover of the cart icon it should also say "Your shopping cart is empty".
6. Click "Continue shopping".
7. This should redirect you to the home page, where you should see the cart icon display the number of items in your cart correctly.

**Expected behavior**
The expectation is that when a user clicks the cart they will be redirected to the checkout view and will successfully see the items they have added to their cart, in addition being able to successfully check out.

**Screenshots**
Screen Shot 2021-01-19 at 4 19 06 PM

 - OS: Catalina
 - Browser: chrome, safari
 - Version 10.15.7 

**Additional context**
This issue has been reported by a few different users with V4.0.0 for example-storefront and V3.11.0 for Reaction.

Akarshit:
Resolves #742 
Impact: **critical**
Type: **bugfix**

## Issue
The cart page was being saved on build time, which was causing the `shop` to null as there was no server to get the shop from during build time.
Also the all the pages were being re-built after 2 mins, because of a wrong variable check.

## Solution
Now the shop would be fetched from the backend on every request. Fixed the variable check to recreate the page when the `shop` is not present.

## Breaking changes
None expected.

## Testing
1. Start the platform using `make`
2. Stop `reaction_api` using `docker stop reaction_api_1`. This would simulate the circle CI build env in which the server is not running.
3. Build the example-storefront image using `docker build --network=host -t  reactioncommerce/example-storefront:test-me .` in `example-storefront` directory
4. Once complete get the server back up using `make init-reaction`
5. Run the new image using `docker run -it --name storefront -p 4000:4000 --env-file .env.prod --network reaction.localhost reactioncommerce/example-storefront:test-me`. If you get an error that `/storefront` is being used by XXX, do `docker rm XXX`
6. Login, try adding products to cart, go to cart page. Everything should work.

There are other pages that use `unstable_revalidate`, but I was getting very sketchy behavior on the first load using that. So I thought it might be better to remove SSR for now, and if needed people can extend and add that according to their need. The behaviour is explained in the linked issue.

### DIFF
--- a/pages/[lang]/cart.js
+++ b/pages/[lang]/cart.js
@@ -16,7 +16,6 @@ import Router from "translations/i18nRouter";
 import PageLoading from "components/PageLoading";
 import { withApollo } from "lib/apollo/withApollo";
 
-import { locales } from "translations/config";
 import fetchPrimaryShop from "staticUtils/shop/fetchPrimaryShop";
 import fetchTranslations from "staticUtils/translations/fetchTranslations";
 
@@ -177,29 +176,17 @@ class CartPage extends Component {
 }
 
 /**
- *  Static props for the cart route
+ *  Server props for the cart route
  *
  * @param {String} lang - the shop's language
  * @returns {Object} props
  */
-export async function getStaticProps({ params: { lang } }) {
+export async function getServerSideProps({ params: { lang } }) {
   return {
     props: {
       ...await fetchPrimaryShop(lang),
       ...await fetchTranslations(lang, ["common"])
     }
-  };
-}
-
-/**
- *  Static paths for the cart route
- *
- * @returns {Object} paths
- */
-export async function getStaticPaths() {
-  return {
-    paths: locales.map((locale) => ({ params: { lang: locale } })),
-    fallback: false
   };
 }
 
--- a/pages/[lang]/index.js
+++ b/pages/[lang]/index.js
@@ -98,7 +98,7 @@ export async function getStaticProps({ params: { lang } }) {
   const primaryShop = await fetchPrimaryShop(lang);
   const translations = await fetchTranslations(lang, ["common"]);
 
-  if (!primaryShop) {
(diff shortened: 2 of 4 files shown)


janus-reith:
I feel like there should be a general solution to this: https://github.com/reactioncommerce/example-storefront/issues/718

Yes, the api would need to be reachable during build time currently.
I tried to point out some potential way to change that in the linked issue.

I doubt however that fully opting out and server-rendering per request instead is a better solution to this.

Akarshit:
The build is failing because of some other reason. I can see that happening in other PRs too.

jrw421:
Hey @Akarshit - I just followed the steps listed to test and still get the same bug. Am I missing something? 🤔

Akarshit:
@jrw421 Hmm, can't say for sure. We can do a quick screen-sharing session to sort this out...

jrw421 (approved):
Tested and looks good! 🎉

PR #751, opened 2021-02-03
Fix: Error on PDP when product is missing slug

### LINKED ISSUE #750: Example-storefront: 404 on PDP when admin saves a product without title
Type: **minor**

**Describe the bug**
Admin can save and publish a product without adding its title. This product, when published, is naturally displayed on the storefront side. When the customer tries to open PDP for that product, the system throws 404 Not Found error (in case of Mac) and Sorry! We couldn't find what you're looking for error (in case of Ubuntu).

**To Reproduce**
1. From reaction admin, create a product without its title (and thus, slug)
2. Publish this product
3. From storefront, open the PDP of the product without title

**Expected behavior**
Error handling on example-storefront for pre-existing products missing titles.

 - MAC OS Catalina
- Ubuntu 20.04 LTS

**Additional context**
There is currently a [PR](https://github.com/reactioncommerce/reaction-admin/pull/340) out for requiring a title on product creation in reaction-admin.

jrw421:
Resolves #750
Impact: **minor**
Type: **bugfix**

## Issue
When a product was saved without a title and published to the storefront and the customer tries to open PDP for that product, the system throws `404 Not Found error` (in case of Mac) and `Sorry! We couldn't find what you're looking for` error (in case of Ubuntu). Related to https://github.com/reactioncommerce/reaction-admin/issues/304. 

## Solution
Only make the link to the PDP available if there is a slug present. If there is a slug, the product links to the product page as expected. Otherwise the product page does not link without a valid slug. If there's a more graceful way of avoiding this error I'm happy to chat through it! 

## Breaking changes
None expected.

## Testing

1. From reaction admin, create a product without its title (and thus, slug)
2. Publish this product
3. From storefront, click on the product without title
4. The click should not link out to the product page or cause the page to error out.
5. Click on a product with a valid slug. 
6. This should bring you to the PDP without issue.

### DIFF
--- a/components/CatalogGridItem/CatalogGridItem.js
+++ b/components/CatalogGridItem/CatalogGridItem.js
@@ -214,7 +214,6 @@ class CatalogGridItem extends Component {
 
   render() {
     const { className, badgeLabels, components: { BadgeOverlay }, product } = this.props;
-
     const { slug } = product;
     const badgeProps = { product };
 
@@ -225,8 +224,8 @@ class CatalogGridItem extends Component {
     return (
       <div className={className}>
         <Link
-          href="/product/[...slugOrId]"
-          as={`/product/${slug}`}
+          href={slug && `/product/[...slugOrId]`}
+          as={slug && `/product/${slug}`}
         >
           <BadgeOverlay {...badgeProps}>
             {this.renderProductMedia()}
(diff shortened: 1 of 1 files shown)


MohanNarayana (changes requested):
The fix works. Please change the commit message format.

MohanNarayana (approved):
Good work

PR #758, opened 2021-03-18
feat: add typescript support

### LINKED ISSUE #709: Begin TypeScript Migration?
# Summary description

Migrate project to TypeScript.

## Implementation

Looks like an easy win as TS can be adopted incrementally. To add support in this project simply:

```sh
yarn add --dev typescript && touch tsconfig.json && dc down && dc up -d
```

Next.js will populate the `tsconfig.json` file with some sane defaults as the container comes up and you can start renaming source files to end with `ts` and `tsx`. (Note at least one TypeScript file is required in the project before the `tsconfig.json` file will be autopopulated with defaults.)

Step two would be to get rid of the `prop-types` dependency. If all goes well, start porting other parts of the framework. Consider pairing this work with https://github.com/reactioncommerce/example-storefront/issues/643.

## Motivation

Type safety. IntelliSense. Better static code analysis resuling in fewer regressions.

## Notes

Doesn't seem to work on the v3.1.0 tag but worked for me on trunk. Probably a Next.js update which allows the `tsconfig.json` file to auto-generate.

ghost:
Resolves #709
Impact: **major**
Type: **feature**

## Issue
DX Suffers without TypeScript

## Solution
Add _strict_ TypeScript support

## Breaking changes
None, though I've changed the code style a little to use functional components, added Prettier to the mix and disabled some ESLint rules in TypeScript files. If we're going to discuss code style now would be a good time. Setting some established patterns now will help others who plan to perform a TypeScript conversion as they fork their Storefronts.

## Configuration

Configure VSCode to use Prettier for TypeScript files by opening the Command Palette (Cmd+Shift+P), opening the Settings JSON file and adding:

```json
"editor.formatOnSave": true,
"[typescript]": {
  "editor.defaultFormatter": "esbenp.prettier-vscode"
},
"[typescriptreact]": {
  "editor.defaultFormatter": "esbenp.prettier-vscode"
},
```

Then run `ext install esbenp.prettier-vscode` to install it.

## Testing
1. Checkout this branch
2. Run `yarn` to install dependencies
2. Run the app
3. Confirm header looks the same as before
4. Run `yarn type-check` from the command line
5. Confirm there are exactly 0 type issues
6. Review `Header.tsx`
7. Confirm it a functional component written in type-script
8. Open the file in VSCode
9. Confirm no ESLint issues appear in the file
10. Observe the `TODO` item for `Link` component
11. Attempt to add a MUI component anywhere in the JSX before it's imported
12. Observe VSCode IntelliSense is working and the `import` statement is added automatically

/cc @janus-reith @nnnnat


types/reaction.d.ts
```diff
@@ -0,0 +1 @@
+declare module "@reactioncommerce/components/ShopLogo/v1";
```
janus-reith, line 1: Do we need this?

components/Header/Header.tsx
```diff
@@ -0,0 +1,90 @@
...
+  shop: {
+    name: string;
+  };
+  uiStore: {
+    toggleMenuDrawerOpen: Function;
+  };
+  viewer: any;
+}
+
+const Header: FC<HeaderProps> = ({ classes, shop, uiStore }) => {
```
janus-reith, line 56: I'd like to discuss the general style we adopt here:
Typescript could infer these, should we utilize that and reduce boilerplate, or do we want to be explicit all the time?
Also, `FC` seems to add no value so we could use `const Header = ({ classes, shop, uiStore }: HeaderProps)` which I would prefer personally.
I don't want to pick on this specific one as it's just an implementation example, but it's style will potentially be adopted in other places aswell.

janus-reith:
Thanks for getting this started @balibebas!

ghost:
@janus-reith Just saw your code review. I'll loop back to this soon.

components/Header/Header.tsx
ghost: My thinking behind use of Type Parameters is to identify type immediately when looking at a function without having parse my eye parse over an argument and possible default params before getting to it, e.g.

```jsx
const Header = ({
  classes: { toolbar, title, controls, appBar },
  shop: { name = "cmVhY3Rpb24vc2hvcDo5TVNZOEZ4ZFI2dTJaOHVNdg==" },
  uiStore,
}: HeaderProps) => {
```

Add an explicit return type, assign default args or define the types inline and it personally starts to give me the bug eyes even after Prettier has gone to work on it, and still suffers from the eye scanning problem I mentioned:

```jsx
const Header = ({
  classes: { toolbar, title, controls, appBar },
  shop: { name = "cmVhY3Rpb24vc2hvcDo5TVNZOEZ4ZFI2dTJaOHVNdg==" },
  uiStore,
}: {
  shop: {
    name: string;
  };
  uiStore: {
    toggleMenuDrawerOpen: Function;
  };
  viewer: any;
}) => {
```

Then there's the moment where you want to return `null` inside an element—likely with a ternary expression—and are required to add `children` to the type definition and you'll still get an error until you choose the correct return type (which `FC` already does for us).

```
Type 'Element | null' is not assignable to type 'HeaderProps'.
  Type 'null' is not assignable to type 'HeaderProps'.ts(2322)
```

The biggest disadvantage I see of using `FC` has to do with potentially adding `children` where we're not dealing with a react component, in which case I prefer the grammar you supplied.

That said it seems as though others prefer omitting `FC` to save a little boilerplate and I'm cool with that so long as we can agree on a way forward to get the migration started and let the style grow out of the refactored code before adding any opinion.

Thoughts?

ghost: For the record, here's the number of times we're returning null in component definitions which can be used as a rough proxy for how often the above error will occur as people are coding: https://github.com/reactioncommerce/example-storefront/search?q=%22return+null%22&type=code

types/reaction.d.ts
ghost: This prevents a type error using modules from the component library. This will eventually become a list of items as they are used from the lib, each added when it becomes used. Not the most straight-forward thing but to me sane until there's a better approach (I haven't found one yet).

Screen Shot 2021-04-01 at 6 46 02 PM

ghost (commented):
@janus-reith all set. once you've considered my responses I'm comfortable moving forward with either grammar for the function signatures so we can get this code merged.

Akarshit:
Let me know if this looks good, and I can get it merged.

package.json
```diff
@@ -34,7 +35,20 @@
...
+    "overrides": [
+      {
+        "files": [
+          "*.ts",
+          "*.tsx"
+        ],
+        "rules": {
+          "comma-dangle": "off",
+          "react/prop-types": "off",
+          "quote-props": "off"
```
ghost: Two more eslint rules worth disabling now as prettier takes care of it:

```json
"react/jsx-max-props-per-line": "off",
"max-len": "off"
```

I'll plan to add these as I incorporate review feedback. 🙏🏼

tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
janus-reith: Should we use "es5" here?
The next's typescript example uses target es next: https://github.com/vercel/next.js/blob/canary/examples/with-typescript/tsconfig.json
and next's itself uses ES2017: https://github.com/vercel/next.js/blob/canary/packages/next/tsconfig.json

TBH I'm unsure what the actual effect of different values here is, since by my understanding, nextjs would not really compile typescript, but typecheck on build and then leverage babel and @babel/preset-typescript to just strip out type definitions.

types/reaction.d.ts
janus-reith: Ok yes I was also expecting this to become a larger list at some point and therefore wondered if we couldn't avoid that somehow. But IMHO not a blocker, something that could easily be refactored later on.

components/Header/Header.tsx
janus-reith: Thanks for taking the time to explain your thoughts on this!
I'm happy to keep it this way.

Longer-term if we deal with passing more data types from the api (which hopefully we won't all type by hand within components anyway and just use codegen) we can still the what patterns fits best, so I'm negating my original statement, not a blocker and nothing that needs to be set in stone right now.

janus-reith (approved):
Looks good to me, thanks @balibebas!

I'd like to reconsider what `target` we set in tsconfig, but approving for now since my by understanding, impact of that property is rather low for us anyways.

Next step might be introducing https://github.com/dotansimha/graphql-code-generator to derive most types from our api schema. By leveraging the typescript-react-apollo plugin and defining some operations (out of the gql files we have already) we can even auto-generate typed apollo hooks for Querys and Mutations and save us a lot of work.

tsconfig.json
ghost: Seems reasonable to shift forward. I suspect it'll reduce bundle sizes by removing boilerplate for things like generator functions where async could be used.

ghost:
@Akarshit we're all set once ci checks have passed. thanks @janus-reith for your feedback and review

nnnnat:
LGTM! If @janus-reith  and @Akarshit are good with this let merge

Akarshit:
I will give this one a last look tomorrow and then merge it.

Akarshit:
I didn't get much time these days, but I will definitely get this done tomorrow.

ghost:
Backing down a change made during review to prevent non-blocking test warning given the TypeScript target `esnext` isn't fully supported by Node 12, the current engine running in the docker container. `es2019` works fine for Node 12.

Prevents warning on Jest test runs:

```
ts-jest[config] (WARN) There is a mismatch between your NodeJs version v12.14.1 and your TypeScript target ESNext. This might lead to some unexpected errors when running tests with `ts-jest`. To fix this, you can check https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping
```

See Node target mappings here: https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping

Akarshit:
@balibebas Hey! looks like to forgot to do the sign-off for the last commit...

ghost:
> @balibebas Hey! looks like to forgot to do the sign-off for the last commit...

Nice catch. Done

balibebas:
Bump to keep the ball rolling on this change.

Akarshit (approved):

Akarshit:
@balibebas Hey! Sorry it was the weekend so I couldn't get to this one. Looks really good to me. Thanks a lot 🎉

PR #760, opened 2021-04-09
feat: use account-js for authentication

Akarshit:
Impact: **major**
Type: **feature**

This PR adds supports to handle authentication via `account-js` and keeps the storefront up-to-date with the latest authentication plugin.

## Breaking changes
This version uses `account-js` to handle authentication. The older accounts created with Hydra/Identity should work. But things like reset password emails that were sent before the update would no longer work.

## Testing
1. Create some accounts with the old hydra/identity setup.
2. Pull this branch, link it as a plugin.
3. Try to access old accounts, those accounts should be accessible.
4. Test signup flow -> click create account, enter email/password, hit Sign Up, this should create an account.
5. Test login flow -> enter email password, the user should be logged in.
6. Test update password flow -> after login, reset password, logout, login with new password.
7. Test forgot password flow -> setup email (or atleast console.log them), try to login, hit Forgot Password, enter email address, get the reset link, click on that, enter new password, try to login with that password


Akarshit:
Updated the PR to pass the builds.

MohanNarayana (approved):

Akarshit:
Only updated the commit message in the last push.

PR #770, opened 2021-06-17
WIP: Add Stripe Payments Intents API SCA-compliant Payment Component (Not intended for merge right now)

### LINKED ISSUE #5304: Refactor the Stripe Card Payment Plugin to use their Payment Intents API and some general considerations
## Issue Description
Currently the Stripe Form executes a call to stripe from the client to create a token:
[/reaction-component-library/src/components/StripePaymentInput/v1/StripePaymentInput.js](https://github.com/reactioncommerce/reaction-component-library/blob/d3ca15bf3c586c8070900d5ddaa3ef088d27be72/package/src/components/StripePaymentInput/v1/StripePaymentInput.js#L78)

These kind of integrations will stop working reliably in the EU from September 14 2019 due to the new [Strong Customer Authentication](https://stripe.com/docs/strong-customer-authentication) rule coming into effect.

The current Token approach is considered outdated, while the new [Payment Intents AP](https://stripe.com/docs/payments/payment-intents/quickstart)I meets the requirement of the SCA.

We should therefore refactor the Stripe payments to this new API.
From what I got so far that kind of integration works similar to the way I did the Paypal integration. The token can be created on the backend and transported to the client using a graphql mutation.

What could be tricky: With Paypal I could trigger such a mutation upon clicking the button, therefore I could always be sure to have the latest total and trigger the checkout from there.

Filling the stripe credit card  form and generating a token doesn't exclude the possibility of changes to the card happening afterwards. While the Payment Intents API takes care of such scenarios by providing an endpoint to update the `amount` of a token, I'd prefer not to make the backend logic too complex just for this provider, or at least see if it makes sense to settle for a generic solution for payment providers with updated cart totals similar to the existing authorize/capture functions on context?

While every payment plugin could just use the cardUpdate appEvents for this currently, a separated function might help to structure things in an easier way?

janus-reith:

Resolves [#5304](https://github.com/reactioncommerce/reaction/issues/5304)
Impact: **breaking|major**
Type: **feature|refactor**

## Issue
The new Stripe Payment api plugin would not be compatible with the existing stripe frontend integration. 

## Solution
This PR adds a new Stripe Credit Card form and utilizes the new mutation to create a Stripe payment intent.

## Breaking changes
This might replace the existing integration in the default storefront, or live side by side with it.
Removing the existing one might be desired though, since it requires an outdated stripe package to be be bundled, as the old integration is not entirely decoupled in the existing component library.

## Testing
See here: https://github.com/reactioncommerce/api-plugin-payments-stripe-sca/pull/2

## Notice & Todo
This branch is intended for testing and should not be merged. I envision this payment component to be distributed as separate module which could be imported from the storefront.
A separate PR might get rid of the existing stripe package included in the storefront, one blocker here is the component library currently throwing errors if that package is not present, regardless of its Stripe Payment Form actually being in use.


delagroove:
merging to another branch in order to fix lint stuff

PR #809, opened 2021-12-20
feat: standardize on one package manager

brent-hoover:
Resolves #None
Impact: **minor**
Type: **refactor**

## Issue

There is no reason to use more than one type of package manager across the project. We use npm in both api and admin so let's use it here, that's one less thing people need to install to get started.

## Solution

Remove the references to yarn from the project

## Breaking changes

As far as I know this should not break anything. You can still use yarn and the development-platform still uses yarn (for now)

## Testing

Everything should work as before


delagroove (approved):

brent-hoover:
Superceeded by #813

PR #810, opened 2021-12-21
chore: make nvmrc consistent

brent-hoover:
Signed-off-by: Brent Hoover <brent@thebuddhalodge.com>

Impact: **minor**
Type: **chore**

## Issue

Different projects are running old non-LTS versions of Node in nvmrc

## Solution

Update it to 14.18.1

## Breaking changes

None

## Testing

Everything should function as before


brent-hoover:
Superceeded by #813

PR #820, opened 2022-04-27
fix: muitheme variable change

delagroove:
Signed-off-by: Alvaro Bueno <alvaro.bueno@mailchimp.com>

fixed an error that shows after changing versions of MaterialUI


vannguyenn (approved):
LGTM

PR #835, opened 2023-03-02
Upgrade node version to 18

sujithvn:
Resolves #6396
Impact: **major**
Type: **feature|chore**

## Issue

Reaction API is being upgraded to Node18. Make the same upgrade for Storefront

## Solution

Updated the Node version in Dockerfile to node:18.10.0-alpine
Node 18 has conflict in crypto/hash [`error:0308010C:digital envelope routines::unsupported`]. The solution/workaround is to use `NODE_OPTIONS=--openssl-legacy-provider`

We have introduced additional commands/scripts in package.json which includes the above mentioned NODE_OPTION as command line parameter. Use the command scripts with `:latest`

*Note*
We have not added assertion for import json as that is handled by webpack (json-loader)

## Breaking changes

Old versions Node 14/15 no longer supported

## Testing
All automated tests should pass
Complete full order placement using Node18 both in Storefront and API


vannguyenn (approved):

sujithvn:
**Note**
We have not added assertion for import json as that is handled by webpack (json-loader)

brent-hoover (approved):

